PR #20, opened 2026-07-10
Fix uvm2020.3.1

Thomasb81:
> **Depends on https://github.com/aionhw/xezim-core/pull/3 and https://github.com/aionhw/xezim/pull/19 . This branch was forked from `nettype`,
   > so its changes appear in this PR's diff until #19 merges. Merge #19 
   > first and the nettype lines drop out automatically.

   ## Summary

   Brings UVM 1800.2-2020.3.1 to parity with UVM 2017. The pipe testbench
   (`GettingVerilatorStartedWithUVM`) deadlocked at `sim_time 0` and ran to
   `max_time`; UVM 2017 completed normally. This PR fixes the four
   SystemVerilog simulation bugs responsible, plus the `+UVM_*` plusargs /
   DPI typedef issues that prevented 2020.3.1 from booting.

   ## Fixes

   1. **`semaphore.try_get` unimplemented (§15.4.3)** — only mailboxes were
      intercepted; semaphore `try_get` fell through and always returned 0,
      so `m_sequence_state_mutex.try_get(1)` failed → `SEQ_NOT_DONE` fatal.
   2. **`if`-condition double-evaluation** — the `If` handler in
      `run_process_stmts` evaluated the condition, then `exec_statement`
      re-evaluated it and the chosen branch. For non-blocking branches the
      body ran twice; `try_get` consumed the key on the first pass, the
      second found it empty. Now evaluates the condition exactly once and
      runs the selected branch inline.
   3. **`m_safe_select_item` not whitelisted** — UVM 2020's forwarding
      wrapper wasn't in `call_is_blocking_task`, so sequencer arbitration
      ran through the synchronous path and couldn't honor `wait`/`@` →
      infinite spin at `t=0`.
   4. **`process::self()` null + fork-child locals unshared (§9.7, §9.3.2)**
      — combined to deadlock the grant handshake. `process::self()` now
      returns an opaque non-null token; fork...`join_none` children now
      *share* automatic variables with the parent scope (§9.3.2) instead of
      a copy, so the parent's `wait` observes the child's write.

   **Boot / plusargs** (`25ff5e2`):
   - `dpi_atom_kind()` resolves `DataType::TypeReference`, so the UVM
     `uvm_hdl_data_t` typedef binds as `svLogicVecVal*` instead of
     `[DPI] unsupported prototype`.
   - Class-body `localparam` (e.g. `localparam string prefix =
     "+uvm_set_verbosity="`) is treated as a static member; previously it
     resolved to `""`, matching every arg → `INVLCMDARGS` / dropped
     `+UVM_TESTNAME`.
   - `uvm_cmdline_processor` now sees the full argv (binary + flags +
     plusargs), matching `uvm_dpi_get_next_arg`.

   ## Results

   - UVM 2020.3.1: **0 errors, 0 fatals, finishes at t=1555** (was: deadlock)
   - UVM 2017: 0 errors, 0 fatals, t=1575 (no regression)
   - Full suite: **470 passed, 0 failed**

   ## Tests

   `dpi_typedef_vec_test` (+ fixtures), `fork_wait_deadlock.sv`,
   `test_fork_join_none_shared_locals` in `sim_tests.rs`, 3 class-localparam
   sim tests.

   ## Note

   Includes a trivial follow-up to `tests/dpi/uvm_dpi_test.c`
   (`cb.user_data = (PLI_BYTE8 *)cookie;`) needed after `main` tightened
   `s_cb_data.user_data` from `void *` to `PLI_BYTE8 *`. A related latent
   `range_select` shift-overflow fix lives in the `xezim-core` repo.


aionhw:
Closing — everything in this PR has landed on main; there's nothing left to merge.

Thanks @Thomas — this PR correctly identified the four bugs behind the UVM 2020 sequencer deadlock at t=0, and its fixes are all on main in one of two forms:

Cherry-picked from this PR (with attribution, in 473b4e5):
- m_safe_select_item added to the blocking-task whitelist, so the UVM 2020 sequencer arbitration chain runs on the suspend-aware path (§9.3)
- process::self() returns a non-null token, so wait(process::self() != null) proceeds (§9.7)
- fork…join_none children share automatic variables back to the parent (§9.3.2), so a child's write is visible to the parent's wait
- Both regression tests (test_fork_join_none_shared_locals, fork_wait_deadlock.sv) were taken as-is

Superseded by independent fixes that landed on main first:
- semaphore try_get/blocking get — fixed in 7b96687 (§15.3.3)
- if-condition double-evaluation — fixed in 68b0444; the conflict was resolved in favor of main's version, which also covers if (e matches p) patterns (§12.6)

With all four in place, UVM 1800.2-2020.3.1 runs green end-to-end (reference testbench: UVM_ERROR : 0 / UVM_FATAL : 0) — see the "UVM library versions" section in docs/uvm-guide.md.

PR #41, opened 2026-07-21
More generous vpi_get_value with ObjectValType and minor cleanup

oscargus:
Try to avoid some of the clippy warnings.

`_ => ` branches can hide bugs (add a new variant and the code compiles).

The removed branches were duplicated. There are more, but I only removed those that I could tell were identical.

Calling `vpi_get_value` with `vpiObjTypeVal` can resolve more types. It is quite convenient to do that for more than the basic types.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -7446,7 +7446,7 @@ impl Simulator {
                         .get(i)
                         .map(|e| self.eval_expr(e))
                         .unwrap_or_else(|| Value::zero(*width));
-                    let (mut aval, mut bval) = Self::dpi_value_to_logic_words(&vv, *width);
+                    let (aval, bval) = Self::dpi_value_to_logic_words(&vv, *width);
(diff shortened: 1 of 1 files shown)


src/compiler/simulator.rs
```diff
@@ -37382,7 +37319,6 @@ impl Simulator {
     /// latent JIT codegen bugs.  Keep signal_has_xz updates limited to
     /// `write_sig!` (full-Value writes) for backward-compatible JIT
     /// behavior.
-    #[inline(always)]
```
oscargus, line 37385: Redundant.

src/compiler/simulator.rs
```diff
@@ -61788,7 +61724,7 @@ union s_vpi_value_union {
 }
 
 #[repr(C)]
-struct s_vpi_value {
+pub struct s_vpi_value {
```
oscargus, line 61727: Warning that this was less visible than a function where it was used.

PR #44, opened 2026-07-21
Uvm genuine fix

Thomasb81:
### Motivation

Bring-up of the UVM class library exposed a cluster of LRM-compliance bugs in three areas that real verification IP exercises heavily: **procedural process control** (§9.3–9.7), **class static-property resolution** (§8.9/§8.25), and **collection-method dispatch** (§8.15/§6.19.6). Each was independently broken and masked by the next. These fixes let the UVM library elaborate and execute its callback, sequencer-handshake, and register-model infrastructure.

---

### Cluster A — Process control & loop semantics (§9.3–9.7)

The suspend-aware statement runner (`run_process_stmts`) correctly handled direct event waits and delays, but missed three transitive-blocking shapes that UVM relies on. Each fix is LRM-correct and pairs with a self-test constructed to fail before the fix and pass after.

| Commit | Fix | LRM |
| --- | --- | --- |
| `53cbb86` | **`break`/`continue`/`return` inside a blocking `forever`** — a `forever` body that can block re-armed on resume by re-appending the `Forever` node, running the body ungated and silently dropping loop-control flags. Now re-appends a `ForeverTail` sentinel (pairs with the xezim-core `ForeverTail` AST variant) so flags are gated on re-entry; first entry stays ungated to avoid stale-flag deadlock. | §9.3.3 |
| `7ad5aed` | **`repeat(N)` with a transitively-blocking body** — `repeat(N) obj.step()` where `step()` contains `wait`/`#delay`/`@event` was not detected as blocking (only direct `@event`/`#delay` triggered unrolling), so the body's task calls were never inlined and a false `wait` silently fell through instead of parking. Mirrors the `while`/`for` arms: also unroll when `stmt_is_blocking(body)`. | §9.4.3 / §9.7.4 |
| `f63903b` | **`process_handle.await()` must block the caller** — `stmt_is_blocking` didn't recognize the await shape, so a `forever { handle.await(); ... }` (the UVM sequencer's re-arbitration fork) busy-spun: the fork child repeatedly re-checked unchanged state, breaking the sequencer handshake. Adds `expr_is_proc_await` static check. | §9.7 |

**New tests:** `forever_break_continue.rs`, `repeat_blocking_call.rs`, `process_class_9_7.rs` (extends existing).

---

### Cluster B — Class static-property & method dispatch (§8.9 / §8.15 / §8.25)

Multi-segment member-access chains starting from a `ClassName::static_prop` head — pervasive in UVM's callback pool (`m_t_inst.m_tw_cb_q`, `m_base_inst.m_pool`) — were resolved incorrectly at every depth: the class name was treated as a null object handle rather than its own class, so reads/writes/method-calls silently no-op'd.

| Commit | Fix |
| --- | --- |
| `036f4c3` | **`ClassName::static_prop` reads/writes/method-calls** — a class name *is* its own class: `ClassName::prop = val` now finds the static cell (was `None`). 2-segment type inference resolves the declared type so `static this_type x = new()` constructs correctly. `ClassName::static_assoc.first(ref k)` returns the collection's bare name rather than treating the class name as a handle. 3-segment method dispatch resolves the static-property handle via `static_prop_key`. |
| `490734b` | **Assoc-array method dispatch + ref-writeback for object keys** — (1) a user method named `num`/`first`/`last`/`next`/`prev` was shadowed by the enum `atoi` fallback, returning 0 for class objects; narrowed the else-branch to numeric-converters only. (2) `infer_width` returned 1 for method formals (`ref widget k`) because it didn't check `local_stack`, truncating object-handle keys to a single bit on writeback (`first(ref k)`/`next(ref k)`). |
| `71b9986` | **Static-property chain reads/writes through object handles** — `ClassName::static_obj.member` type-walk and `= val` writes now resolve the static head to its handle and route the leaf to the shared static cell. New `member_handle`/`read_member_value` helpers fall back to the instance's *class* static cell when a property isn't in per-instance storage (§8.25: a static accessed through an instance refers to the shared static). Applied consistently to `eval_handle_expr`, `eval_expr`, and `assign_value`. |

**New tests:** `static_property_access.rs`, `assoc_method_dispatch.rs`.

---

### Cluster C — Test infrastructure

| Commit | Change |
| --- | --- |
| `bd0128a` | **Parser-gap self-tests** — pure-SV tests for the xezim-core parser fixes they pair with: `disable iff` in default clocking (§16.6), `virtual`-interface type argument (§25.9), inherited-type queue port `name [$]` (§6.3). New file: `parser_gaps2.rs`. |
| `6ba57bb` | **`RUST_MIN_STACK=32M` via `.cargo/config.toml`** — the simulator evaluates deeply recursive SV constructs (concat chains, nested member access, parameterized class construction) whose debug-build frames can exceed Rust's default 2 MB test-thread stack, aborting test binaries before they emit a result line. Pre-existing (confirmed against pre-merge code); release builds unaffected. Harmless for the main binary. |


.cargo/config.toml
Thomasb81, line 1: Not sure about this change : impact testsuite in debug mode

PR #54, opened 2026-07-23
Install polly in CI

oscargus:
Should hopefully fix the compilation error.

I dropped the cache as the disk was full earlier. Now there is a different error, but not sure if it is related or not.

Edit: seems like the error is also out of disk (or RAM). Maybe one should split in three different jobs, for now I'm testing to remove the build files from the previous tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,16 +16,6 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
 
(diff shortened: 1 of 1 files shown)


aionhw:
Obsoleted by the LLVM JIT backend removal (31e26fe, 8f6973a) — no LLVM/jit-llvm path remains, so Polly is moot and re-adding the jit-llvm CI job would fail with 'unknown feature jit-llvm'.

Issue #11, opened 2026-07-09
Support for Arrays as Parameters


Ganesh-TNL:
Sample test case on EDA Playground:

https://www.edaplayground.com/x/a3kA

Simulation log with Aldec Riviera Pro in the above link shows the logic working as expected for a 'selection BFM' where the number of select pins as well as the address range for each are parameterizable.

xezim 0.8.2 has a parser error for the same code.

```
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim arr_param.sv
=== xezim ===
Max time: 100000
------------------------------
Simulation error: Parse errors in source 0:
error: expected RParen, found Identifier 'START_ADDR' (at byte 182..192)
error: expected Semicolon, found Identifier 'START_ADDR' (at byte 182..192)
error: expected Colon, found RBracket ']' (at byte 204..205)
error: expected expression, found RBracket ']' (at byte 204..205)
error: expected RBracket, found Assign '=' (at byte 206..207)
error: expected identifier, found Assign '=' (at byte 206..207)
error: expected Semicolon, found Identifier 'END_ADDR' (at byte 223..231)
error: expected Colon, found RBracket ']' (at byte 243..244)
error: expected expression, found RBracket ']' (at byte 243..244)
error: expected RBracket, found Assign '=' (at byte 245..246)
error: expected identifier, found Assign '=' (at byte 245..246)
error: expected Semicolon, found RParen ')' (at byte 262..263)
error: unexpected: ")" (at byte 262..263)
error: unexpected: "(" (at byte 264..265)
error: expected identifier, found KwInput 'input' (at byte 287..292)
error: expected Semicolon, found KwInput 'input' (at byte 287..292)
error: expected identifier, found KwOutput 'output' (at byte 307..313)
error: expected Semicolon, found KwOutput 'output' (at byte 307..313)
error: expected Semicolon, found RParen ')' (at byte 346..347)
error: unexpected: ")" (at byte 346..347)
```

PR #55, opened 2026-07-24
Uvm genuine fix

Thomasb81:
## Summary

   This branch implements the parameterized-class specialization handling
   that UVM's callback subsystem depends on, plus a fix for a concurrent
   local-collection stall. All changes are LRM-grounded, each with a
   pure-SystemVerilog self-test, and each fix was validated byte-for-byte
   against reference simulators.

   `main` is fully merged in (merge-base == `main` HEAD), so this is
   fast-forwardable with no conflicts.

   - **18 commits** (12 work + 6 `main` merges)
   - **12 files changed, +2058 / −96**
   - **10 new self-test files**; **1174 self-tests pass, 0 failures**
     (13 ignored — the deferred queue-local-isolation cases)

   ---

   ## 1. UVM callback machinery: parameterized-class statics

   UVM's `uvm_callbacks#(T,CB)` builds a typewide callback graph out of
   static singletons and typeid maps inherited across parameterized-class
   hierarchies. Several specialization gaps left that graph empty or
   corrupted:

   - **Typedef static scope, per-spec static keying, type-param locals**
     (`1d91861`) — a plain `typedef base super_type` used as a static-call
     scope (`super_type::m_initialize()`) was silently dropped, so the
     shared callback pool never initialized. Inherited statics
     (`uvm_typed_callbacks#(T)::m_tw_cb_q`) now share one cell across all
     CB specializations of the same T via `ancestor_spec`, and
     type-parameter-typed locals in **static** methods now resolve through
     `current_spec` rather than the empty `this_stack`. LRM §6.18/§8.25.1.

   - **`$cast` on a type-parameter-typed destination** (`9da190c`) —
     `$cast(me, obj)` where `me` is `T me;` must check against T's
     concrete specialization, not a too-loose ancestor. The recorded type
     was the param *name* ("T"), so `cast_type_ok` fell into the permissive
     unknown-type branch and every such cast succeeded — leaking a
     typewide callback for one specialization onto unrelated instances.
     LRM §6.20.2/§8.25.

   - **Specialization-signature canonicalization** (`099267e`) — `C#(arg)`
     (defaulted trailing param omitted) and a typedef-based
     `C#(T,ST,CB)` reference must resolve to the same static cell.
     `canonicalize_spec_sig` substitutes defaulted params and pads missing
     trailing positions, so inherited statics like
     `uvm_callbacks::m_typeid` are written and read under one key
     (depends on the `type_param_defaults` field in the xezim-core PR).
     LRM §6.20.2.

   - **Virtual-dispatch spec restore on sibling specs** (`7cbd0e1`) — the
     dispatch spec-restore only switched `current_spec` when the class
     *name* differed, so a virtual call between
     `uvm_callbacks#(base)` and `uvm_callbacks#(derived)` (same base,
     different specialization) left `current_spec` stale and typewide
     callbacks never propagated to derived types. Also resolves an elided
     trailing defaulted type param to its default. Self-test:
     `tests/typeparam_default_resolution.rs`.

   - **Class-property value-param binding + `$cast` value-param specs**
     (`f6920cb`) — property construction (`this.prop = new(...)`) now binds
     value params from the property's recorded `#(args)`, and `$cast`
     across value-parameter specializations now compares value params
     (after stripping `#(args)` for the class-hierarchy check). Self-tests:
     `tests/property_value_param_binding.rs`,
     `tests/cast_value_param_specs.rs`.

   - **Class-property type over stale same-named local** (`0ae7364`) —
     `get_expr_type_name` trusted the flat `var_class_types` accumulator
     (never scope-cleared) over class properties, so a local from an
     *earlier* method leaked in and shadowed a same-named property during
     its `new()`. Now the flat maps are trusted only for genuine locals of
     the current scope.

   - **Per-specialization static storage** (`8f2c99c`) — static
     initializers for parameterized classes now run once per
     specialization (via `ensure_spec_statics`, triggered at construction
     and first parameterized static-method dispatch) with `current_spec`
     set, instead of once for the generic class; and static **collection**
     members (queue / dynamic / associative arrays) are now keyed by the
     spec-aware `C#spec::member` instead of a bare member name, so
     `C#(int)::q` and `C#(bit)::q` no longer share element/size cells.
     Self-test: `tests/static_collection_per_spec.rs`.

   ## 2. Per-process local-collection storage (time-0 stall)

   - **Local associative-array isolation** (`02f2470`) — automatic
     task/method locals are per-invocation (LRM §6.21/§9.3.2). Two
     concurrent fork/join_none siblings each declaring a local associative
     array previously shared one `name.size` / `name[i]` cell in the global
     signals table, the second `VarDecl` clobbering the first. Each local
     assoc-array declaration now gets a process-unique key (`@name#id`)
     recorded in a per-call-frame rename map. This was the root cause of a
     time-0 stall where concurrent `uvm_phase_hopper::sync_phase()`
     invocations clobbered each other's local `edges_t edges`. Verified
     byte-for-byte. Queue/dynamic-array local isolation is implemented but
     deferred (it regresses a register-model whole-queue-copy path); its
     self-tests are `#[ignore]`'d behind a `|| true` guard.

   - **Queue-local module registration restore** (`8dbed02`) — the per-frame
     save/restore snapshotted a shadowed queue's `signals` but not its
     `module.arrays` / `dynamic_arrays` / `widths` registration, so after a
     nested same-name local returned the caller's queue stayed
     unregistered: later `p.delete()` was a silent no-op and `p[i]` read
     empty. `QueueLocalSave` now captures and restores the full
     registration. Self-test: `tests/queue_local_shadow_restore.rs`.

   ## 3. Repo hygiene

   - `bcfde6a`, `977b549` — remove accidentally-tracked build artifacts
     and probe scratch.
   - `bf103d7` — untrack local helper/probe files (kept on disk).

   ## Test status

   - **1174 / 1174** self-tests pass, 0 failures (13 ignored — deferred
     queue-local-isolation cases).
   - All fixes validated byte-for-byte against reference simulators.


Thomasb81:
Rely on https://github.com/aionhw/xezim-core/pull/11

Issue #12, opened 2026-07-09
Support for Array of Queues


Ganesh-TNL:
Sample test case on EDA Playground:

https://www.edaplayground.com/x/Mfvi

Simulation log with Aldec Riviera Pro gives the following correct display outputs:

```
# KERNEL: === Starting Array of Queues Verification ===
# KERNEL: 
# KERNEL: --- 1D Array of Queues Contents ---
# KERNEL: queue_array_1d[0] size = 2, contents = '{10, 11}
# KERNEL: queue_array_1d[1] size = 1, contents = '{20}
# KERNEL: queue_array_1d[2] size = 3, contents = '{30, 31, 32}
# KERNEL: 
# KERNEL: --- 2D Array of Queues Contents ---
# KERNEL: queue_array_2d[0][0] size = 2, contents = '{99, 100}
# KERNEL: queue_array_2d[0][1] size = 2, contents = '{2, 101}
# KERNEL: queue_array_2d[0][2] size = 2, contents = '{3, 102}
# KERNEL: queue_array_2d[1][0] size = 2, contents = '{2, 110}
# KERNEL: queue_array_2d[1][1] size = 2, contents = '{99, 111}
# KERNEL: queue_array_2d[1][2] size = 2, contents = '{4, 112}
# KERNEL: 
# KERNEL: Popped from queue_array_2d[0][0]: 99
# KERNEL: 
# KERNEL: === Verification Passed Successfully ===
```

xezim 0.8.2 seems to have an issue all through for the same piece of code, with the assertions / queue functions appearing to not work correctly as well.

```
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim array_of_queues.sv
=== xezim ===
Max time: 100000
------------------------------
[PHASE] compilation: 0.8ms
[EVENT-EDGE] measure (timestamp): 0 edge blocks, 0 gateable, 0 gateable-with-EMPTY-data-reads, avg data-reads/gateable=0.00
[EVENT-EDGE] SKIP mode ON (sim_time >= 0) — skipping gateable flop fires with no data-input change
=== Starting Array of Queues Verification ===

--- 1D Array of Queues Contents ---
queue_array_1d[0] size = 0, contents = x
queue_array_1d[1] size = 0, contents = x
queue_array_1d[2] size = 0, contents = x

--- 2D Array of Queues Contents ---
queue_array_2d[0][x] size = 0, contents = x
[...]

PR #57, opened 2026-07-27
Uvm genuine fix

Thomasb81:
## Summary

   Follow-up runtime fixes for correct parameterized-class specialization,
   constructor dispatch, and UVM methodology support. Depends on the
   companion  https://github.com/aionhw/xezim-core/pull/12 PR (parser/elaboration/value).
   +1988 / −139 across 21 files, 21 commits.

   ## Changes

   ### Parameterized-class specialization & `$cast`
   - Enforce `$cast` across value-param specializations via type-parameter
     destinations: `class_of_var` returns the specialization form instead
     of discarding it, `cast_type_ok` passes spec forms through, and static
     properties are read from the shared static cell, not a stale
     per-instance copy.
   - Preserve and canonicalize spec context during static typedef
     instantiation; don't set `instance.spec` unconditionally at
     construction (fixes a `uvm_config_db` / `uvm_resource_db`
     set/get disagreement regression).
   - Don't treat body-`localparam` classes as parameterized for static init
     (§6.20.2), so their static initializers run (eliminates the CBUNREG
     callback-registration barrier).

   ### Constructor `new(...)` dispatch
   - Never dispatch `new()` as a static method (was constructing onto a null
     handle, producing ghost components and infinite recursion in phase
     traversal); `new()` inside a static method now allocates a real instance.
   - Resolve `prop = new(...)` over the property's declared type, not a
     same-named caller's leaked local (fixes a stack overflow).
   - Resolve value-parameter references in specialization arguments and add
     a cycle guard in `resolve_value_param_from_spec` (fixes another stack
     overflow).
   - Construct class-property initializers of the form `T x = new(...)`
     (§8.7/8.12) that the constructor never reassigns.
   - Constructor arg dispatch, struct-field queue assignment, typedef
     resolution in the `is_new_call` guard, and queue isolation.

   ### Class-field named events
   - A class `event` field now has backing storage: `@field` blocks (instead
     of returning immediately at t=0) and `->field` wakes the waiters. Adds
     a per-instance named-event waiter list mirroring module-scope events;
     parked processes count as suspended for `fork...join_any` / `disable
     fork` correctness (underpins `uvm_event` / `uvm_heartbeat`).

   ### Value semantics
   - Empty string literal `""` has width 0, not 8 (§6.16), so comparisons
     against zero-width strings no longer widen and return X.

   ### UVM runtime phase scheduling
   - Enable the full sequential phase schedule (pre_reset → reset → … →
     post_shutdown): handle nested member-access (`a.b.c`) for queue
     properties in the objection drain path, and stop the
     `end_run_phase` shim from setting `finished=true` prematurely in
     PURE_SV_LRM mode.


aionhw:
Thanks for this — it's a substantial, well-decomposed set of fixes, and the
per-fix self-tests (13 new test files, +1247 lines) make it reviewable one
commit at a time. I validated it against current main rather than reading
only: it merges cleanly, builds clean, and the full suite passes at 274
targets / 1275 tests / 0 failures with this PR, the companion
xezim-core#12, and my in-flight parameterized-class work all applied together.

Issue #13, opened 2026-07-09
Support for Associative Arrays


Ganesh-TNL:
Sample test case on EDA Playground:

https://www.edaplayground.com/x/KG35

It is a simple dual 'bank' memory model where the 'address' is used as a hash key (so that the simulator only allocates memory for written addresses) - a concept adopted to be able to simulate huge amounts of memory attached to a system.

Aldec Riviera Pro has no problem in filling up and retrieving the written content.

xezim 0.8.2 retrieves either 0 or x.

```
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim assoc_arr_case.sv
=== xezim ===
Max time: 100000
------------------------------
[PHASE] compilation: 0.5ms
[EVENT-EDGE] measure (timestamp): 0 edge blocks, 0 gateable, 0 gateable-with-EMPTY-data-reads, avg data-reads/gateable=0.00
[EVENT-EDGE] SKIP mode ON (sim_time >= 0) — skipping gateable flop fires with no data-input change
Retrieved entry in mem 0 addr 04 : 0000000x
Retrieved entry in mem 0 addr 05 : 0000000x
Retrieved entry in mem 0 addr 06 : 0000000x
Retrieved entry in mem 0 addr 07 : 0000000x
Retrieved entry in mem 0 addr 40 : 00000000
Retrieved entry in mem 0 addr 41 : 00000000
Retrieved entry in mem 0 addr 42 : 00000000
Retrieved entry in mem 0 addr 43 : 00000000
Retrieved entry in mem 0 addr 44 : 00000000
Retrieved entry in mem 0 addr 45 : 00000000
Retrieved entry in mem 0 addr 46 : 00000000
Retrieved entry in mem 0 addr 47 : 00000000
Retrieved entry in mem 0 addr 48 : 00000000
Retrieved entry in mem 1 addr 04 : 0000000x
Retrieved entry in mem 1 addr 05 : 0000000x
Retrieved entry in mem 1 addr 06 : 0000000x
Retrieved entry in mem 1 addr 07 : 0000000x
Retrieved entry in mem 1 addr 40 : 00000000
Retrieved entry in mem 1 addr 41 : 00000000
Retrieved entry in mem 1 addr 42 : 00000000
Retrieved entry in mem 1 addr 43 : 00000000
Retrieved entry in mem 1 addr 44 : 00000000
Retrieved entry in mem 1 addr 45 : 00000000
[...]

PR #58, opened 2026-07-29
Uvm genuine fix

Thomasb81:
12 simulator fixes for SystemVerilog (IEEE 1800) language semantics, each validated byte-for-byte against the reference simulator.

 Class generics & type parameters (§8.25)
 - Resolve type parameter as receiver of a class-scoped static property read (T::static_prop)
 - Resolve type-parameter receiver in ClassName::type_id::create(...) specialization
 - Specialization-aware static collection keying & dispatch for parameterized classes (pkg::cls#(T, CB))

 Class member semantics
 - Apply queue/dynamic-array class member default initializers (int q[$] = '{1,2,3}) and preserve queue contents on method return
 - Dispatch chained size()/len() on a class handle returned by a call to the user-defined method
 - Bind unpacked struct value arguments to subroutines member-wise

 Type reflection
 - $typename for parameterized class handles (§21.7)
 - Enum .name() reflection on formal method/function parameters, with package-prefix resolution fallback

 Operators & expressions
 - String ==/!= is 2-state (§6.16); integral ==/!= on mixed types remains 4-state
 - Sign-extend narrow RHS on sized local assignment (§10.7)
 - Offset vector bit/range select by the declared lower bound (§7.4.1 / §11.5.1)

 File I/O
 - MCD/FD split model for $fopen/$fwrite/$fclose (§21.3.1)


aionhw:
Thanks for this — a lot of real ground covered, and the test files are welcome. I reviewed it merged onto current `main`, and reference-compared every finding below.

## This cannot be merged on its own

#58's **own** tests `queue_member_default_initializer` and `assign_from_queue_member` fail against current `main` — 927 passed, **2 failed**. The queue-initializer half needs `elaborate.rs`, which lives in `xezim-core`, and this PR only touches the `xezim` repo.

The other half is **aionhw/xezim-core#13**. Merged as a pair the suite is green: **300 targets, 1378 passed, 0 failed**. Same pairing as core#12 + xezim#57 last round — worth noting in the PR description so the two don't get separated.

## Blockers

### 1. Leftover debug print

`ensure_spec_statics` ships an unconditional write to stderr:

```rust
self.class_statics.insert(spec_key.clone(), v.clone());
eprintln!("[DBG_SPEC_INIT] spec_key={} val={:?}", spec_key, v);
```

A two-specialization toy program already emits:

```
[DBG_SPEC_INIT] spec_key=holder#int,4::stamp val=Value { storage: Inline { val_bits: 1, xz_bits: 0 }, width: 32, ... }
[DBG_SPEC_INIT] spec_key=holder#bit,8::stamp val=Value { storage: Inline { val_bits: 2, xz_bits: 0 }, width: 32, ... }
```

On a UVM run this is continuous.

### 2. Regression: a method-local that shadows a string property is printed as text

`class_member_is_string` matches a bare `Ident` against `this`'s properties **without first checking whether a local of that name is in scope**:

```systemverilog
class C;
  string val = "abc";
  function void show();
    int val;  val = 65;          // method-local shadows the property
    $display("SHADOWBARE", val);
  endfunction
endclass
```

| | output |
|---|---|
| reference simulator | `SHADOWBARE         65` |
| xezim `main` | `SHADOWBARE65` |
| **this PR** | **`SHADOWBAREA`** |

65 comes out as the character `'A'`.

The underlying concern is scope: `expr_is_string_valued` has four call sites, and widening it now also affects `$display`'s "leading string argument is the format spec" path and the concatenation text-join — a larger blast radius than the string `==`/`!=` change it was written for. A local-scope check in `class_member_is_string` before consulting `this` should cover it, but the other call sites are worth a look too.

### 3. Stray file at the repo root

`test_param_cb_isolation.sv` (73 lines) is added at the top level rather than under `tests/`.

## Should fix

### 4. The lower-bound select fix is read-only

For `logic [7:4] v` / `logic [3:1] w`:

| | `main` | this PR | reference |
|---|---|---|---|
| read `v[6:5]` | 0 | **1** ✓ | 1 |
| write `w[3:2] = 2'b11` | 4 | **4** ✗ | 6 |

The three changed sites are all read paths; `assign_value` never gets the same offset, so writing a part-select of a non-zero-LSB vector still lands on the wrong bits. (Bit-selects were already correct in both directions, so the gap is specific to part-select writes.)

Minor, same area: the new sites clamp with `.max(0)`, so an out-of-range select reads bit 0 rather than `x`.

### 5. core#13's gate removal is broader than the simulator handles

Dropping `decl.dimensions.is_empty()` records initializers for **every** dimensioned member, but the simulator side only special-cases queues and dynamic arrays:

| member | `main` | this PR | reference |
|---|---|---|---|
| `int q[$] = '{10,20,30}` → `size()` | 0 | **3** ✓ | 3 |
| `int d[] = '{7,8}` → `size()` | 0 | **2** ✓ | 2 |
| `int fa[4] = '{1,2,3,4}` → `fa[0]` | 0 | 0 ✗ | 1 |

Not a regression — `main` is 0 too — but fixed unpacked arrays now fall through to `inst.properties.insert(pname, val)`, storing an assignment pattern as a scalar in the scalar property map. Either handle them alongside queues, or keep the gate and add an explicit dimensioned-member branch.

## Design concerns (not blocking, but they will compound)

### 6. `uvm_pkg::` is hardcoded into general simulator paths

The literal prefix now appears in `get_class_def`, `class_is_a`, `enum_value_name`, `bind_unpacked_struct_arg`, `class_declared_type_of`, `class_prop_struct` and others, and `normalize_spec_ws` does `s.replace("uvm_pkg::", "")` on *every* specialization signature.

The real problem being solved is package-qualified class-name resolution. As written it does nothing for any other package, and each new lookup path will need the same two-line fallback appended. A single normalization helper (or resolving package-qualified names once at elaboration) would fix the whole class of cases instead.

### 7. Stall detection is silently gated to `depth == 0`

```rust
if self.stall_limit > 0 && depth == 0 {
```

Re-entrant activations no longer count toward the livelock limit. That may well be correct — a recursive call shouldn't inflate a per-pid counter — but it isn't mentioned in the PR description and it weakens a guard that has caught real bugs. Worth a comment explaining the reasoning.

### 8. `cast_type_ok` churn drops explanatory comments

```rust
- if !self.class_is_a(&src_class, &resolved) {
+ let is_a = self.class_is_a(&src_class, &resolved);
+ if !is_a {
```

and the same for `cast_value_params_ok`. No behaviour change, but it deletes ~30 lines of comments explaining why the `$cast` permissiveness and value-parameter checks exist. Those came out of earlier debugging; could they be restored?

## Credit where it's due

The MCD/FD model (§21.3.1) is the strongest part of this PR — correct bit-31 discrimination, disjoint internal keying for the two ranges in one table, stdout/stderr handled inline so they are never closed, and genuinely helpful comments. The queue/dynamic-array member init and return fixes, chained `size()`/`len()` dispatch on a returned handle, and the read-side select fix are all real and well covered by the new tests.

## Suggested path

Items 1–3 before merge, then land #58 and xezim-core#13 **together** (or neither). 4 and 5 are fine as follow-ups as long as they're tracked. Happy to re-run the suite and the reference comparison once you've pushed.

Thomasb81:
A batch of 21 commits hardening the SystemVerilog simulation runtime (src/compiler/simulator.rs) for UVM correctness, plus 22 new/updated self-tests (+3647 / −188 lines). The
 recurring theme is UVM class/factory semantics, but the fixes are all general SystemVerilog-language correctness fixes validated against a reference simulator on pure-SV
 reproducers (byte-for-byte on assertion lines). Grouped by theme:

 ### Class factory, specialization & type parameters

 - $typename for parameterized class handles (§21.7) — was returning "logic" for every class-typed expression; now renders "class <name>#(<args>)" with recursive type-arg
   rendering.
 - Type-parameter receiver in T::type_id::create(...) — the interception that constructs classes directly had no path for a type parameter receiver (the pattern inside every
   parameterized UVM class's type_name()).
 - Type parameter on static-property reads (T::static_prop) — was silently reading zero because the read path only consulted module.classes.
 - Type-check copy-ctor arg against declared type — T x = new(arg) where arg is an unrelated class was misclassified as the §8.12 shallow-copy form, skipping the constructor
   body (broke the UVM factory instance-override path).
 - Static callback collection keying for parameterized classes — spec-aware keying in Foreach resolution + canonicalization of names like uvm_callbacks#(T, CB).
 - Refactor: drop redundant uvm_pkg:: class-name fallbacks — the elaborator now canonicalizes package-qualified names, so the simulator's per-site format!("uvm_pkg::{}", …)
   probing is removed.

 ### Strings & enum reflection

 - String ==/!= is 2-state (§6.16) — routing string equality through the 4-state integral path returned X on X-padding; now both-string-operand comparisons yield a definite
   bit. Plus sign-extend narrow RHS on local assignment (§10.7).
 - Class methods named like §6.16 string builtins not shadowed — compare/tolower/atoi/etc. were dispatched to the string builtin on any receiver; now guarded by
   class_expr_has_method().
 - Enum .name() reflection on formal method parameters + extended to free functions — register formal-parameter types on entry; package-prefix fallback in enum_value_name.

 ### Collections (arrays, queues, structs)

 - Unpacked array equality for class fixed-array properties (§11.4.5) — rhs_.sa == this.sa fell through to scalar comparison in do_compare; new fixed_array_operand() resolves
   whole-array operands (module-level, this.member, obj.member, flattened).
 - Class member queue initializers & queue returns — int q[$] = '{1,2,3} was ignored; return <queue_member> lost contents. New populate_queue_from_init + instance_assoc_member
    resolution.
 - No-parens .size/.num on class collection properties (§7.5.2/§7.9.1/§7.10.2.1) — accept the MemberAccess form as a constraint operand and resolve instance-scoped storage.
 - Chained size()/len() on a returned class handle dispatches to the real method instead of the string-byte-length fallback.
 - Bind unpacked struct arguments member-wise — struct value args (incl. string fields) were unbound in subroutine scope.
 - Size function return variable to declared return-type width (§13.4) — the implicit return cell was hardcoded 32-bit and not in the width table, dropping bits ≥32 (broke
   wide UVM pack/unpack).

 ### Bit/range selects & part-select writes

 - Vector bit/range select for non-zero lower bounds (§7.4.1/§11.5.1) — offset select indices by the declared lower bound (reg [11:-4] w; w[8:0]).
 - Part-select writes apply the declared lower-bound offset (in the #58 cleanup commit) — logic [3:1] w; w[3:2] = 2'b11 now yields 6, matching the reference simulator.

 ### File I/O & config_db

 - MCD/FD file I/O model for $fopen/$fwrite/$fclose (§21.3) — split the 32-bit file value by bit 31: MCD (bit 31 clear, multi-channel) vs FD (bit 31 set, single descriptor).
   UVM writes through UVM_STDOUT = 32'h8000_0001 (an FD) previously vanished.
 - Intercept uvm_config_db#(T) static method calls — set/get/exists were bypassing xezim's config-DB scope matching; now intercepted with a non-re-entrant internal helper.

 ### Multi-top support & review cleanup

 - Multi-top-module self-tests (§23.3.3) — self-tests for the xezim-core multi-top elaboration fix; relax one timescale assertion under the __xezim_multi_top wrapper.
 - Review blockers on #58 — remove debug eprintln!, fix local-shadowing-string mis-typing, initialize fixed-size unpacked array class members member-wise, add
   class_storage_key() normalization, restore dropped comments.

PR #66, opened 2026-07-31
Uvm genuine fix

Thomasb81:
This set of changes targets a cluster of SystemVerilog/UVM semantics in the
   simulation runtime, with emphasis on the UVM factory/callback machinery,
   method dispatch, and I/O that the UVM library depends on. Each fix is pinned
   by a new self-checking self-test whose SystemVerilog was validated
   byte-for-byte against a reference simulator on the assertion lines.

   ## Generics, type parameters & the UVM factory

   - **`$typename` for parameterized class handles** (§21.7) — renders
     `class <name>#(<args>)` instead of `"logic"`; recurses on type args.
   - **Resolve a type parameter on static-property reads** (`T::static_prop`),
     matching the existing method/typedef dispatch.
   - **`T::type_id::create(...)` resolves a type-parameter receiver** — the
     pattern inside every parameterized UVM class's `type_name()`; previously
     returned null, producing factory type names like `"foo #()"`.
   - **Static callback collection keying & dispatch for parameterized
     classes** (e.g. `uvm_callbacks#(T,CB)`).
   - **Intercept `uvm_config_db` static method calls.**
   - **Refactor:** drop the redundant `uvm_pkg::` class-name fallbacks now
     that the parser/elaborator canonicalizes package-qualified names upstream.

   ## Functions, methods & constructors

   - **Size the function return variable to the declared return-type width**
     (§13.4) — the implicit return cell was hardcoded 32-bit, so bit-select
     writes `funcname[i]=...` for `i >= 32` (UVM's 64-bit `uvm_integral_t`
     pack/unpack core) silently dropped the upper half.
   - **Type-check the copy-ctor argument against the declared type** (§8.12) —
     `T x = new(arg)` with `arg` an *unrelated* class handle is no longer
     misclassified as the shallow-copy form `new src`; unblocks the UVM
     factory instance-override path.
   - **Bind unpacked-struct subroutine arguments member-wise.**
   - **Dispatch chained `size()`/`len()` on a returned class handle to its
     real method** — the byte-length fallback was shadowing user methods on
     handles.
   - **Class methods named like the §6.16 string builtins no longer
     shadowed** (`tolower`/`toupper`/`atoi`/…): only dispatch the builtin
     when the receiver's type has no such method.
   - **No-parens `.size`/`.num` on class collection properties.**
   - **Formal-parameter type registration** for free functions; **enum
     `.name()` reflection on formal method parameters.**
   - **Queue member default initializers** (`int q[$] = '{1,2,3}`) are now
     applied, and **returning a queue member** preserves its contents.

   ## Expression / type system

   - **String `==`/`!=` is 2-state** (§6.16) — comparisons of the `string`
     data type yield a definite 0/1, never X from capacity padding; mixed
     `byte != "."` stays integral.
   - **Sign-extend narrow RHS on local assignment** (§10.7) —
     `logic [63:0] mask = -1` yields all-ones inside class methods.
   - **Unpacked-array equality compares element-by-element** (§11.4.5) — now
     works for class fixed-array properties (`obj.sa == this.sa`), not just
     module-level arrays.

   ## Lvalues, indices & selects

   - **Vector bit/range select for signals with non-zero or negative lower
     bounds** (§7.4.1/§11.5.1) — offsets select indices by the declared
     lower bound.
   - **Part-select WRITES apply the declared lower-bound offset** (matches the
     read path) — `logic [3:1] w; w[3:2] = 2'b11` yields 6, not 4.
   - **Fixed-size unpacked array class member initializers** applied
     member-wise.

   ## I/O

   - **MCD/FD file-I/O model for `$fopen`/`$fwrite`/`$fclose`** (§21.3) — bit 31
     distinguishes multichannel descriptors from file descriptors; UVM writes
     through `UVM_STDOUT = 32'h8000_0001` (an FD) were previously dropped.

   ## Elaboration / tests

   - **Multi-top-module self-tests** (§23.3.3) for the elaboration fix that
     lives in xezim-core; relax one timescale assertion for the multi-top
     wrapper.


aionhw:
Reviewed merged onto current `main` and paired with aionhw/xezim-core#13; full suites run on the pair, and every finding below was reference-compared on minimal reproducers.

**Suite: 326 targets, 1480 passed, 3 failed — all 3 in one new test file** (`tests/uvm_config_db_tests.rs`). Core 33/0, parser 54/0.

All #58 blockers are properly addressed — the debug print is gone, the string-shadow regression now matches the reference (`SHADOWBARE65`), the stray root file is gone, and the flagged comment deletions became a real refactor. Thanks for the fast turnaround on those.

## Blocker: `tests/uvm_config_db_tests.rs`

1. Three of its four tests shell out to a **hardcoded path in your home directory** (`/home/tom/prog/git/xezim/xezim/target/debug/xezim`) — they fail with "No such file or directory" on any other machine, which is exactly the 3 suite failures.
2. `run_questa` invokes a commercial simulator by name (36 mentions) from the committed suite. This repo keeps vendor names out of the tree entirely — reference comparison belongs in the dev workflow, not in committed tests.
3. The fix is mechanical: rewrite on the in-process `simulate()` API like the other 23 test files (the fourth test in the file already works that way).

## Newly wrong: the lower-bound offset also hits indexed selects' WIDTH bound

The part-select write fix (thanks for taking review item 4) subtracts the declared lower bound from **both** bounds, but in `[i +: W]` / `[i -: W]` the right bound is a **width**, not an index. For `logic [3:1] w` / `logic [7:4] v`:

| | this PR | reference |
|---|---|---|
| `w[3:2] = 2'b11` | 6 ✓ | 6 |
| `v[5 +: 2]` read | `10` | **`01`** |
| `w[3 -: 2] = 2'b00` | `011` | **`001`** |
| `w[1 +: 2] = 2'b11` | `001` | **`011`** |

Offset the base index only; leave the width alone. (Current `main` is also wrong on the indexed forms — differently — so this PR is strictly closer, but as written the Constant-range fix trades one wrong shape for another.)

## Newly wrong: fixed-array member initializer non-positional forms

`int fa[4] = '{4{9}}` evaluates the replication as one 128-bit value and stores it in `fa[0]` (`713053462794…`, rest 0); `'{default:7}` fills only element 0. Reference: `9 9 9 9` / `7 7 7`. The positional path (including non-zero LSB `fa[1:4]`) is correct. Routing items through the same expansion `pattern_ordered` uses plus a default fill would cover it — or bail to the old all-zeros behavior for non-positional items rather than storing a garbage packed value.

## Notes, not blockers

- **Copy-ctor type-check (§8.12)**: good, well-commented, and the right call for unrelated types. The residual case is a **same-type** single ctor argument: `b = new(a)` where the class declares `function new(T parent)` — still shallow-copies; the reference calls the constructor (`tag=42`, parent set). The LRM distinguishes `new src` from `new(arg)` *syntactically*, so the durable fix is preserving that shape at parse time. Fine as a tracked follow-up since the conflation predates this PR.
- **config_db interception**: works, but it repurposes `pure_sv_lrm` as a reentrancy guard around the nested UVM call — that flag gates many unrelated behaviors, so anything inside `set()` that consults it changes semantics during the window. You already added `exec_static_method_internal`; a dedicated guard bool would make the flag flip unnecessary.
- The **MCD/FD model**, the multi-top pair, chained `size()`/`len()` dispatch, `$typename`, and the string-builtin shadow fix all verified clean against the reference here.

## Merge logistics

Must land **together with aionhw/xezim-core#13 or not at all** (the queue-init tests need both sides). Also heads-up: `main` moved this week (part-select bounds, class-member signedness, construction-time init stamping all touch your hunks) — I hit 3 conflicts merging for review; a rebase will need care specifically in `assign_value`'s RangeSelect arm and the `property_inits` loop.

**Verdict:** with the config_db test file rewritten (in-process, vendor-free), the indexed-select offset corrected, and the array-init non-positional forms fixed or made benign, this is mergeable as a pair — the batch is genuinely good work, and the test discipline (20 new self-checking files) is exactly what this repo wants.

Thomasb81:
Issue addressed

aionhw (approved):
Re-reviewed at fa84f24 merged onto current `main`, as a pair with aionhw/xezim-core#13.

**Approve.** All three blockers from the last round are properly resolved:

1. **`tests/uvm_config_db_tests.rs`** — rewritten on the in-process `simulate_multi` API: no subprocess, no hardcoded home-directory path, zero vendor-simulator mentions. With the Accellera 1800.2-2020.3.1 reference implementation checked out as a sibling (`../1800.2-2020.3.1`, `accellera-official/uvm-core` tag `2020.3.1`), all 3 tests pass here.
2. **Indexed part-select lower-bound offset** — re-verified against a commercial reference simulator on `logic [3:1]` / `logic [7:4]` covering `[b +: W]` / `[b -: W]` reads and writes: 5/5 lines byte-identical. Offsetting only the base index (and both bounds only for the constant `[msb:lsb]` form) is exactly right.
3. **Fixed-array member initializer forms** — `'{4{9}}` and `'{default:7}` now route through `pattern_elems`; verified 9 9 9 9 / 7 7 7 7 with a non-zero lower bound.

**Pair suite: 337 targets, 1521 passed, 0 failed** — fully green.

One advisory (non-blocking) follow-up: when `../1800.2-2020.3.1` is absent the three tests *panic* rather than skip, so a fresh clone without the UVM checkout sees 3 failures. A skip-guard (return early with an eprintln when the file is missing — same pattern the binary-harness tests use) or a README note on the expected sibling checkout would make the suite self-contained again. Fine as a follow-up commit.

Thanks for the thorough turnaround — the reference-verified commit message made re-review fast.

Issue #14, opened 2026-07-10
Support for Bidirectional Pass Switches


Ganesh-TNL:
Sample test case on EDA Playground:

https://www.edaplayground.com/x/eRzK

This tests compliance with Section 28.8  of IEEE Std 1800-2017. Successful processing log with Aldec Riviera Pro is available in the link above.

xezim 0.8.3 output log:

```
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim bidir-switch-compliance-00.sv
=== xezim 0.8.3 ===
Max time: 100000
------------------------------
[PHASE] compilation: 1.2ms
[EVENT-EDGE] measure (timestamp): 0 edge blocks, 0 gateable, 0 gateable-with-EMPTY-data-reads, avg data-reads/gateable=0.00
[EVENT-EDGE] SKIP mode ON (sim_time >= 0) — skipping gateable flop fires with no data-input change
==========================================================
Starting SystemVerilog LRM Section 28.8 Compliance Test
==========================================================
[ERROR] TEST 1A: tran (Left=1, Right=Z) Failed!
        Left Net  -> Expected: 1, Got: 1
        Right Net -> Expected: 1, Got: x
[PASS] TEST 1B: tran (Left=Z, Right=0) validated successfully.
[ERROR] TEST 1C: tran Contention (Left=1, Right=0) Failed!
        Left Net  -> Expected: x, Got: 1
        Right Net -> Expected: x, Got: 0
[ERROR] TEST 2A: tranif1 Disabled (Ctrl=0) Failed!
        Left Net  -> Expected: 1, Got: z
        Right Net -> Expected: 0, Got: 0
[ERROR] TEST 2B: tranif1 Enabled (Ctrl=1) Failed!
        Left Net  -> Expected: x, Got: 0
        Right Net -> Expected: x, Got: 0
[PASS] TEST 2C: tranif1 Pass-through (Ctrl=1) validated successfully.
[ERROR] TEST 3A: tranif0 Disabled (Ctrl=1) Failed!
        Left Net  -> Expected: 0, Got: z
        Right Net -> Expected: 1, Got: 1
[ERROR] TEST 3B: tranif0 Enabled (Ctrl=0) Failed!
        Left Net  -> Expected: x, Got: 1
        Right Net -> Expected: x, Got: 1
[ERROR] TEST 4A: tranif1 Unknown Control (Ctrl=X) Failed!
[...]

PR #71, opened 2026-08-01
Add self-test for macro token pasting (IEEE 1800-2023 §22.5.1)

### LINKED ISSUE #69: Double-backtick macro re-expansion does not work
### Description

Double-backtick macro parameter re-expansion (IEEE 1800-2017 section 22.5.1) doesn't work in xezim. The `` \` ``\` ``  token is ignored and the formal parameter name is emitted as an undeclared identifier.

This breaks many Questa-originating macro patterns that use `` \` ``\` `` to re-expand arguments.

### Minimal repro

```sv
`define HELLO "world"
`define WRAP(v) ``v``

module repro_double_btick;
   initial begin
      $display("HELLO=%s", `HELLO);            // works: prints "world"
      $display("WRAP(HELLO)=%s", `WRAP(HELLO)); // should print "world" but errors
      $finish;
   end
endmodule
```

### Expected

Both $display calls should print `"world"`. The double-backtick in `` `WRAP(HELLO) `` should expand HELLO to `"world"` and re-scan.

### Actual

```
$ xezim --simulate -s repro_double_btick repro.sv
Simulation error: Undeclared identifier 'HELLO'
```

### Notes

- Single-backtick `` `MACRO `` expansion works correctly
- The double-backtick `` \` ``\`formal``\` ``\`  `` construct is ignored, causing the raw formal name to appear in the source
- This is used heavily in testbench macro wrappers (e.g. UVM sequence macros)
- xezim git: `3053d6d`

Thomasb81:
Test for https://github.com/aionhw/xezim-core/pull/14 to fix  https://github.com/aionhw/xezim/issues/69

### DIFF
--- a/tests/preprocessor_github_issues.rs
+++ b/tests/preprocessor_github_issues.rs
@@ -133,6 +133,17 @@ fn double_backtick_outside_a_string_joins_tokens() {
     assert_eq!(squash(&out), "foo_bar", "got {out:?}");
 }
 
+/// IEEE 1800-2023 §22.5.1: Token pasting `` `` `` occurs before macro lookup on the resulting merged token.
+#[test]
+fn double_backtick_token_pasting_before_macro_lookup() {
+    let out = pp("`define FOO_BAR 42\n`define CONCAT(x) `FOO_``x\n`CONCAT(BAR)\n");
+    assert_eq!(squash(&out), "42", "got {out:?}");
+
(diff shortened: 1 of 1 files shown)


aionhw (approved):
**Approve** — reviewed together with aionhw/xezim-core#14 (full review there). Both tests encode the reference-validated semantics, including the important one: `` `PASTE(A,B) `` = 99 proves paste-before-lookup rather than expand-then-paste. Suite with the pair merged: 315/1473/0.

PR #74, opened 2026-08-02
fix(simulator): dispatch §6.16 string methods on hierarchical receivers

### LINKED ISSUE #72: string.len() returns 0 for interface/hierarchical string signals
## Steps to reproduce

```systemverilog
interface my_if;
    string data = "hello_world_this_string_is_valid";
endinterface

module consumer;
    my_if u_if();
    initial begin
        $display("value  = '%s'", u_if.data);
        $display(".len() = %0d  (expected 33)", u_if.data.len());
    end
endmodule

module top;
    consumer u_consumer();
endmodule
```

Save as `repro.sv`, run with:
```
xezim --simulate -s top --max-time 1us repro.sv
```

## Actual output

```
value  = 'hello_world_this_string_is_valid'
.len() = 0   (expected 33)
```

## Expected behavior

`u_if.data.len()` should return `33` (the string length). The string value is accessible and prints correctly via `%s`, but `.len()` returns 0.

## Impact

This causes `$sformatf("%s", hier_string)` and `{s1, s2}` concatenation to produce wrong-length results, corrupting any string built from interface/hierarchical `string` signals. In practice this truncates long constructed strings to 128 characters, breaking `$system()` commands built via `$sformatf` + `{...}` concatenation.

## Version

xezim 0.9.7 (git 15e7a3d), debug build

Thomasb81:
fix: https://github.com/aionhw/xezim/issues/72

IEEE 1800-2023 §23.6 (hierarchical names) + §6.16 (string methods). A string field reached through a hierarchical reference — e.g. `u_if.data.len()` inside a `consumer` module — returned 0 (and `u_if.data.getc(1)` returned 0, `u_if.data.substr(1,3)` returned "").

Root cause: the parser flattens `a.b.m()` into `Ident([a, b, m])`, and the generic hierarchical-call tail derived the receiver as `path[0]` (`a` = `u_if`) instead of the full `path[0..len-1]` (`a.b` = `u_if.data`). For the common 2-segment `localvar.m()` case `path[0]` and the joined receiver are identical, so it worked; any 3+-segment hierarchical string receiver lost its middle segment. The §6.16 builtins then ran against the interface instance name (`u_if`), which has no value, and returned 0.

The string methods were also excluded from every prior Ident-arm handler: `is_array_builtin_method`'s collection block (a string field is not a collection) and the no-arg string block (its `args.is_empty()` gate excludes `getc`/`substr`).

Fix: a dedicated block for `len`/`size`/`getc`/`substr`/`compare`/ `icompare`/`atoreal` on a hierarchical-ident receiver. It reconstructs the receiver `Ident(path[..len-1])`, resolves its fully-qualified name (`resolve_hier_name` → `u_consumer.u_if.data`), and dispatches the builtin by that name — the handlers read the value via `get_local_signal_or_static`, which finds the hierarchical signal. A USER-METHOD guard (evaluate the receiver to a handle; if its class defines the method, fall through to real method dispatch) preserves precedence: a class with its own `len()` still calls the real method, not the byte-length fallback.

Validated byte-for-byte against the reference simulator (len/getc/ substr/tolower all match). All three self-tests fail without the fix (len=0, getc=0) and pass with it; the user-method-precedence test passes both ways. No regressions in the self-test suite (the single pre-existing failure, test_pr_pr1990029, fails on clean main too).

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -64355,6 +64355,70 @@ impl Simulator {
                 }
             }
 
+            // §6.16 string methods (len/size/getc/substr) and §6.16.6
+            // (compare/icompare/atoreal) on a hierarchical-ident receiver
+            // (`u_if.data.len()`, `top.dut.cfg.substr(0,3)`). They were
+            // excluded from `is_array_builtin_method`'s collection block
+            // above (a string field is not a collection) and from the
+            // no-arg string block below (its `args.is_empty()` gate), so
+            // they fell through to the generic tail, which mis-derived the
+            // receiver as just `path[0]` instead of the full
+            // `path[0..len-1]` (IEEE 1800-2023 §23.6 — the receiver of
+            // `a.b.m()` is `a.b`, not `a`; the parser flattens the call
+            // into `Ident([a, b, m])`).
+            //
+            // `len`/`size`/`getc`/`substr` dispatch by the resolved
+            // hierarchical name (their handlers read the value via
+            // `get_local_signal_or_static`, which finds the hierarchical
+            // signal); `compare`/`icompare`/`atoreal` evaluate the receiver
+            // by value. The USER-METHOD guard is essential: a class instance
+            // whose class defines a same-named method (e.g. a `container`
+            // with its own `len()`) must dispatch to the real method, not
+            // the string builtin — without it `c.len()` on a container
+            // local returned the byte length of the handle instead of
+            // calling `container::len`. The guard evaluates the receiver to
+            // a handle and asks `class_has_method` (value-based, mirroring
+            // the MemberAccess `len`/`size` fallback).
+            if len >= 2
+                && matches!(
+                    path[len - 1].name.name.as_str(),
+                    "len" | "size" | "getc" | "substr"
+                        | "compare" | "icompare" | "atoreal"
+                )
+            {
+                let m = path[len - 1].name.name.clone();
+                let base = HierarchicalIdentifier {
+                    root: hier.root.clone(),
+                    path: path[..len - 1].to_vec(),
+                    span: hier.span,
+                    cached_signal_id: std::cell::Cell::new(None),
+                    cached_resolved_name: std::cell::OnceCell::new(),
+                };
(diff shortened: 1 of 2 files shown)


aionhw (approved):
**Approve.** Reviewed merged onto current `main`; the diagnosis in the PR body matches what I see, and the behavior is byte-identical to a commercial reference simulator on the repro and on a wider probe.

The root cause is exactly as described: the parser flattens `a.b.m()` into `Ident([a, b, m])` and the generic tail took the receiver as `path[0]`, which is only correct for the 2-segment case. Verified against the reference:

| case | before | after / reference |
|---|---|---|
| `u_mid.u_if.data.len()` (3-segment) | 0 | 10 |
| `u_mid.local_s.len()` (2-segment) | 12 | 12 |
| `u_mid.u_if.data.getc(1)` | 0 | 98 |
| `u_mid.u_if.data.substr(2,4)` | `""` | `"cde"` |

The user-method guard is the part I looked hardest at, and it is right to be there: with a `container` class defining its own `len()`/`size()`, `c.len()` still dispatches to the class method (99/77, not the handle's byte length). Worth keeping the comment that explains why.

Suite with the PR merged: **345 targets, 1547 passed, 0 failed** (+3 from the PR's own tests). Full ivtest sweep: no change, no regressions.

One note for the issue, not the PR: the reporter's "expected 33" is off by one — `hello_world_this_string_is_valid` is 32 characters, and 32 is what both this fix and the reference produce.

Thanks — clean diagnosis, minimal patch, and the test file covers the shapes that matter.

Thomasb81:
Superseded by #75, which includes this fix as its final commit along with the broader batch of dispatch/construction/builtin fixes.

aionhw:
Merged as 2df3941 (`Merge pull request #74 from Thomasb81/string-hier-methods`), commit 770b7ac is on `main`.

Note GitHub shows this as **closed** rather than **merged**: its mergeability check for the PR was stuck reporting `unknown` and the merge API refused, so I merged the branch locally and pushed. The content is identical to what the button would have produced (a `--no-ff` merge commit) — nothing was rebased or squashed, and your commit keeps its authorship.

Verified on `main` after the merge: your three tests pass, and the full suite is green.

PR #75, opened 2026-08-02
Class dispatch, construction, and builtin-method fixes

Thomasb81:
A batch of correctness and performance fixes for class method dispatch, class-member construction, and built-in method handling. Each fix is covered by self-checking tests; the full self-test suite passes (1560/0). No behavioral change for designs that did not exercise the affected paths.

## Performance (class dispatch and type resolution)

- `exec_method_in_class_hierarchy` now clones only the single matched method body when walking the inheritance chain, not the whole class (was O(class_size) per call; pathological under deep inheritance and heavy call load).
- Width inference skips collection-method detection for callees that are not builtin names.
- `resolve_type_param_with` stops cloning the entire class.

Test: `tests/class_method_dispatch.rs`

## Class-member construction and return-type sizing

- The implicit function return variable is registered at its declared return-type width, initialized to the type default, and its declared class/typedef type is registered so a bare `funcname = new(...)` resolves the class to construct.
- Struct member assignments and property access work for structs whose fields are class handles.
- Associative-array elements whose element type is a typedef'd class are constructed with the resolved class.
- Class property initializers using the bare `new` syntax (no argument list) now construct the instance.

Tests: `tests/class_handle_return_preservation.rs`, `tests/struct_with_class_handle.rs`, `tests/assoc_typedef_element_class.rs`, `tests/property_bare_new_initializer.rs`

## Static method binding and type-parameter specialization (§8.20, §8.25)

- A non-virtual method binds to the receiver's declared type, not the runtime type. Unqualified non-virtual and local method calls now resolve statically through the declared class's hierarchy.
- A type parameter specialized to a specialized class name binds the type parameter to that specialization.

Tests: `tests/local_method_static_binding.rs`, `tests/type_param_bound_to_specialization.rs`

## Built-in methods on hierarchical and assoc-of-queue receivers (§6.16, §7.9.1, §12.7.3)

- §6.16 string methods (len/size/getc/substr/compare/icompare/atoreal) on a hierarchical-ident receiver such as `u_if.data.len()`. A user-defined class method of the same name still takes precedence over the string builtin.
- Associative-array-of-queue top-level key enumeration (num/first/last/next/prev) and two-variable `foreach(all[k,i])` over an assoc-of-queue now iterate the sparse outer keys correctly.

Tests: `tests/hierarchical_string_method.rs`, `tests/assoc_of_queue_enumeration.rs`

---

**Note:** Supersedes #74 (that standalone fix is now the final commit here).


aionhw (approved):
**Approve.** Reviewed merged onto current `main` (which already carries #74 as 2df3941 — your final commit merges cleanly as a no-op). Every claim below was checked against a commercial reference simulator on probes I wrote independently of your tests, and each behavioral change was also baselined on `main` to separate genuine fixes from pre-existing gaps.

**Gates**
* Suite with the PR merged: **353 targets, 1559 passed, 0 failed** (+8 test files).
* Full ivtest sweep (921 designs): **no change** — nothing gained, nothing regressed.
* Perf claim confirmed on a class-dispatch-heavy workload (`bench/crosstool/b4_oop_tb`, small preset): **1.94 s → 1.46 s, ~25 % faster**, checksum unchanged.

**Verified fixes** (each broken on `main`, correct here, matching the reference):

| behavior | main | this PR | ref |
|---|---|---|---|
| §8.20 non-virtual binds to declared type (`b.who()` / `b.vwho()` / inner call) | — | 1 / 20 / 120 | same |
| `funcname = new(args)` return handle, plain **and** typedef'd return type | 0 / 0 | 5 / 6 | same |
| bare-`new` property initializer | — | 7 | same |
| struct with class-handle member, **automatic local** | 0 | 9 | 9 |
| assoc-of-queue: `num`/`first`/two-var `foreach` | — | 3 / 1 / 10 / 4 | same |

**Non-blocking scope notes.** Three claims in the description read more broadly than what actually landed. All three are *pre-existing* gaps (identical on `main`), so nothing here regresses — but I'd rather the PR text matched, and these are good follow-up candidates:

1. *"Struct member assignments and property access work for structs whose fields are class handles"* — holds for an `automatic` local (your test's shape). A **module-scope** `wrap_t w; w.h = new(9); w.h.v` still reads **0** (reference: 9).
2. *"Associative-array elements whose element type is a typedef'd class are constructed"* — holds for a class **property** assigned with bare `new` (your test's shape). A module-scope or automatic-local `item_t aa[string]; aa["k"] = new(11); aa["k"].v` still reads **0** (reference: 11).
3. *"A type parameter specialized to a specialized class name binds the type parameter to that specialization"* — holds for an **inline** specialization (`registry#(wrap#(int))`). Bound through a **typedef alias** (`typedef payload#(32) p32_t; user#(p32_t)`) it still reads **0** (reference: 32).

Suggestion: soften those three bullets to name the shape each covers, or open issues for the remaining ones so they are not lost. Either is fine by me; I'm not holding the merge on it.

One structural remark: at 11 commits and ~1 900 changed lines in `simulator.rs`, this is at the outer edge of what I can review as a unit. The commits are individually clean and well-scoped, which is what made it tractable — if a future batch grows past this, splitting the perf commits from the semantic ones would keep bisection meaningful. Thanks for keeping each fix paired with a self-checking test.

aionhw:
Follow-up to my review, now that this is merged (thanks!). I left the three scope gaps as prose, which is a good way to lose them — here they are as concrete, reference-validated cases so they can be picked up (or split into issues) later. All three are **pre-existing**, not caused by this PR: each behaves identically on the commit before it.

Each snippet was run against a commercial reference simulator; `xezim` column is current `main` (9ad0d3f).

**1. Struct with a class-handle member at MODULE scope**

```systemverilog
class item; int v; function new(int x = 7); v = x; endfunction endclass
typedef struct { item h; int tag; } wrap_t;
module tb;
  wrap_t w;                       // module scope (not `automatic`)
  int c;
  initial begin w.h = new(9); c = w.h.v; $display("C %0d", c); end
endmodule
```
xezim `C 0` — reference `C 9`. The same code with `automatic wrap_t w;` inside the initial block gives 9 on both, which is the shape `tests/struct_with_class_handle.rs` covers.

**2. Assoc array of a typedef'd class, outside a class property**

```systemverilog
class item; int v; function new(int x = 7); v = x; endfunction endclass
typedef item item_t;
module tb;
  item_t aa [string];
  int e;
  initial begin aa["k"] = new(11); e = aa["k"].v; $display("E %0d", e); end
endmodule
```
xezim `E 0` — reference `E 11`. Fails at module scope and as an `automatic` local. `tests/assoc_typedef_element_class.rs` covers the collection-as-class-property form assigned with bare `new`, which does work.

**3. Type parameter bound through a TYPEDEF ALIAS of a specialization**

```systemverilog
class payload #(parameter int W = 8);
  function int width(); return W; endfunction
endclass
typedef payload#(32) payload32_t;
class user #(parameter type T = payload#(8));
  T inst;
  function new(); inst = new(); endfunction
  function int w(); return inst.width(); endfunction
endclass
module tb;
  user #(payload32_t) u32;   // alias, not an inline specialization
  int a;
  initial begin u32 = new(); a = u32.w(); $display("A %0d", a); end
endmodule
```
xezim `A 0` — reference `A 32`. The inline form `user #(payload#(32))` works, as does `registry#(wrap#(int))` in `tests/type_param_bound_to_specialization.rs`.

The common thread in 1 and 2 is that the fix keyed off the *storage class* of the collection (automatic local / class property) rather than the element type, so module-scope declarations miss it; 3 is the typedef alias not being resolved to its underlying specialization before binding. No action needed from you on this thread — recording it so the gaps are findable.

Issue #73, opened 2026-08-01
PURE_SV_LRM=0 question


Thomasb81:
Hello

Does the undocumented mode `PURE_SV_LRM=0`  need to be maintained ?

Initial UVM support rely on this mode which provide a rust implementation of UVM internal instead of a systemVerilog. This have the benefice to speed-up UVM support in the early days at the price of a missing part of the systemVerilog LRM.  

BR
Thomas

aionhw:
that switch will be removed permanently. still fixing other issues and will  take care that switch afterward.

aionhw:
Good question, and the short answer is: **no, it does not need to be maintained as a feature — but please do not remove it in the same change that touches anything else.**

Where things stand today:

* Pure-LRM is the **default** (`pure_sv_lrm` defaults to true; `PURE_SV_LRM=0` opts back into the legacy mode).
* The legacy mode still gates roughly 16 UVM-specific shim sites in the simulator.
* Exactly **one** committed test still runs with `PURE_SV_LRM=0` (`tests/uvm_integration_tests.rs`, the TLM `write_in()` case), plus the mock header that keeps `uses_real_uvm()` true for it.

So the mode has already served its purpose: it bootstrapped UVM support before the genuine library ran, and the genuine library now runs. My preference for retiring it:

1. **Do not delete it as part of another PR.** Every one of those 16 shim sites is a behavioral fork; folding them in alongside an unrelated fix makes a regression impossible to bisect.
2. **Retire it in its own change**, one shim at a time where practical, each with the pure-LRM path demonstrated to cover the case the shim existed for. The single `PURE_SV_LRM=0` test should be converted to the genuine path (or deleted with a note saying what it used to cover) rather than left behind as the last user of a mode nobody else exercises.
3. **Keep the env var accepted-but-ignored for one release** if you want a soft landing for anyone scripting it, then drop it.

If you would like to take that on, I would review it — a PR that removes the flag and the shims, with the suite and an ivtest sweep clean, is a change I would be happy to merge. What I would push back on is a partial removal that leaves some shims gated and others not, because then neither mode is really tested.

Thanks for asking before touching it.

Thomasb81:
##  The problem (current architecture)

Built-in methods are dispatched through three independent mechanisms today:

| Mechanism | Dispatch basis |
|---|---|
| `eval_builtin_method` | **name-based** switch over `string`/`queue`/`assoc`/`event`/`enum` |
| `string_method` | **value-based**, string only |
[...]

PR #79, opened 2026-08-02
Retire Hardcoded UVM Shims & Enable Native SystemVerilog LRM Execution

Thomasb81:
Related to #73

  ## Motivation & Summary

  This PR retires the legacy PURE_SV_LRM=0 mock/shim mode and removes all hardcoded simulator-level UVM interceptors, special-cased method dispatches, and synthetic phase backstops.

  With this refactoring, Accellera UVM (IEEE 1800.2-2020.3.1) now runs natively via standard SystemVerilog semantics—relying on native class inheritance, method dispatch, process control,
  associative arrays, and DPI bindings rather than simulator-level shims.

  ## Key Changes

  ### 1. Removal of Hardcoded UVM Method Dispatches & Interceptors

  • Method Dispatch & Calls: Removed hardcoded simulator routing for run_test, uvm_root singleton dispatch, config_db set/get virtual-interface interception, and factory::create shims. All
  subroutines and method dispatches now execute natively.
  • Objection & Phasing Engine: Removed 11 simulator-side tracking fields and interceptor backstops (e.g., uvm_obj_count, uvm_obj_raised, uvm_phase_drain, uvm_pending_end,
  handle_uvm_objection, end_run_phase, run_uvm_cleanup_phases, bridge_objection_wait_for). The Accellera UVM phase scheduler, objection tracking, and callbacks now execute natively in SV.
  • CLI & Helper Interceptors: Removed simulator-level uvm_glob_match, uvm_dpi_get_next_arg, and uvm_cmdline_processor hooks in favor of standard DPI-C and SV execution.

  ### 2. Method Dispatch & Process Control Fixes

  • Fixed method-chain call resolution and procedural loop stalling when executing deep object hierarchies natively.
  • Standardized process continuation chaining (pc.pushed(...)) for inlined static methods and procedural loops to ensure blocking tasks yield correctly without spinning.

  ### 3. Cleanup of Legacy Mocks & Tests

  • Removed legacy mock headers (tests/uvm/uvm_mock.svh) and mock integration tests (tests/uvm_integration_tests.rs).
  • Removed outdated PURE_SV_LRM=0 mode flags and updated doc comments across the codebase.
  • Code Reduction: Net reduction of >2,200 lines of code in src/compiler/simulator.rs (-3,301 lines, +757 lines).

  ## Verification & Testing

  • Self-test Suite: Added tests/pure_sv_phase_objection.rs and tests/procedural_loop_stall.rs. Full self-test suite is 100% clean (cargo test passes all tests).
  • Reference Simulators: All fixed semantics verified byte-for-byte against reference simulators.


aionhw:
Reviewed by applying this branch on top of current `main` (3-way apply, no conflicts, builds clean) and running the full self-test suite.

## Blocking: 7 tests regress

`tests/udp_primitives.rs` goes from **14 passed / 0 failed** on `main` to **9 passed / 5 failed** with this branch applied. The UDP tests do not mis-compare — they produce **no output at all**:

```
UDP trace mismatch for `mux`:
got:
      <empty>
expected:
  t=0 sel=0 a=0 b=0 q=0
  t=1 sel=0 a=1 b=0 q=1
  ...
```

Failing: `comb_mux`, `comb_and_xor`, `level_latch`, `instance_delay`,
`primitive_verbose_is_scoped_to_explicit_v_files`, `v_file_and_y_dir_resolve_on_demand`,
`fscanf_advances_only_past_what_it_matched`.

## Why I think that happens — the flag removal is doing more than shim removal

`main` has **42 references to `pure_sv_lrm`, of which 32 are conditionals**. Deleting the flag turns each one into an unconditional decision in the *default* execution path, and several of those are unrelated to UVM. Two I spotted directly in the diff:

```diff
-  && (self.stmt_is_blocking(s) || (self.pure_sv_lrm && Self::stmt_is_mailbox_get(s)))
+  && (self.stmt_is_blocking(s) || Self::stmt_is_mailbox_get(s))
```
a mailbox `get` now always counts as blocking, and

```diff
-  let unknown_typeref_handle = self.pure_sv_lrm && !two_state
+  let unknown_typeref_handle = !two_state
```
a bare `TypeReference` that names neither a known width-typedef nor a known class now defaults to `null` instead of `x` for every design, not just in pure mode.

Each of those 32 sites is a real semantic choice that deserves its own justification; folded into a 2,700-line deletion they are invisible.

## Also

The PR deletes `tests/uvm_integration_tests.rs` and `tests/uvm/uvm_mock.svh` — the coverage for exactly the paths being removed — and replaces them with two much smaller tests. That leaves little evidence that native UVM still works after the change.

## Suggestion

Split this into (a) removal of the hardcoded interceptors, and (b) the 32 gate decisions, ideally in small batches with the reasoning for each direction. The direction is right and the code reduction is welcome; it is the bundling that makes it unreviewable and, per the suite, currently incorrect.

Happy to re-run the full suite on any follow-up push.

Thomasb81:
Mark PR as draft for reference. will come back with a re-work PR.

Thomasb81:
Superseded by #87 — same work re-applied onto current `main` as a single clean commit, addressing the review above. The 7 regressions were stale-base interaction artifacts; on a fresh rebase they're all gone and the full suite (1048 tests) passes at the clean-`main` baseline. Closing this one; further review welcome on #87.

Thomasb81:
Closing in favor of #87, which supersedes this branch with a clean rebase onto current `main` (no stale-base interaction artifacts). See the cross-reference comment above.

PR #82, opened 2026-08-03
fix(simulator): unpacked-struct class properties read/write/compare

Thomasb81:
## Problem

   Whole-value access of an unpacked-struct **class property** — a class field
   whose type is an unpacked struct — was broken:

   ```systemverilog
   typedef struct { int a; int b; } s_t;
   class pair #(type T1 = int, type T2 = int);
     T1 first;   // <-- unpacked-struct class property
     T2 second;
     ...
   endclass
   ```

   Internally, an unpacked-struct property is stored **member-wise** as
   `<prop>.<member>` cells in the instance heap (`first.a`, `first.b`, …). There
   is no single container cell for `first`. Consequently:

   - a **whole read** of `first` (e.g. `if (first == rhs.first)` inside a method,
     or `%p` / `convert2string`) returned a stale/zero container value;
   - a **whole write** of `first` (e.g. `c.first = s0;` outside the class, or
     `first = rhs.first;` inside `copy()`) was silently lost;
   - **`==`/`!=`** between two such properties (or a property vs. a top-level
     struct variable) always evaluated to `X`.

   Field-level access (`c.first.a`) already worked; only the *whole-value* view
   was missing.

   ## Fix

   Bridge the whole-value view by **reusing the already-correct field-level
   read/write paths**, rather than materializing the struct as a single value.
   This keeps the same heap representation and only adds an assembly/decomposition
   layer. (IEEE 1800-2023 §7.2, §7.3.1, §10.6, §11.4.5, §18.4.)

   **Write** — `try_decompose_struct_class_prop_assign`:
   A whole-struct assignment whose target is an unpacked-struct class property is
   decomposed into per-member field assignments:

   ```
   c.first = s0   →   c.first.a = s0.a;  c.first.b = s0.b;
   ```

   `append_member_expr` builds each field expression by appending a member segment
   (to a hierarchical `Ident` path, or wrapping in `MemberAccess`). Each side is
   then read/written by its working field path, so the RHS is read member-by-member
   and never needs a (broken) whole-struct value. Nested struct/array members
   recurse naturally through the same path.

   **Read** — `read_whole_struct_class_prop` / `try_read_whole_struct_class_prop`:
   Assemble the `<prop>.<member>` cells into one packed integral blob (first member
   at the MSB, matching SV packed layout) using the existing `struct_field_layout`.
   This is hooked into all three parse shapes that can name a whole property:

   - implicit-`this` bare name (`first` inside a method),
   - 2-segment flat hierarchical ident (`c.first`),
   - `MemberAccess` form (`rhs.first`),

   as well as the class-handle-deref walk, which now tracks `cur_handle` so a
   property reached through a handle chain is assembled from the **current** object.

   **Compare** — `compare_unpacked_structs` + `class_prop_struct_of`:
   If *either* operand resolves to a class-property struct (members in the heap),
   compare member-wise by appending a member segment to each side and evaluating
   through the field paths — uniform for property-vs-property and
   property-vs-top-level-signal. The pre-existing signal-namespace leaf-flatten
   path is left untouched for plain signal structs (so nested-struct/string
   members and `X`-propagation behavior there are unaffected).


Thomasb81:
Scope expanded — now 3 commits

 The branch grew from the class-property fix to cover passing structs through
 function/method boundaries. All three share this PR's member-wise heap representation

 2. Type-parameter struct formals (74b20d9)
 A struct formal typed as a class type parameter (function new(T f) where T is a struct)
 was never bound member-wise — the body read every member as x. bind_unpacked_struct_arg
 resolved the type name against the typedef table, but a type parameter isn't a typedef.
 Fix: resolve the type parameter to its concrete bound type first (via the instance's
 type_bindings, or current_spec for constructors), mirroring the existing
 dimension-resolution path.

 3. output/inout/ref struct formals (e9ae324)
 An output/inout/ref struct formal was bound member-wise but never written back — the
 caller's actual stayed x. Affects all struct types. Two gaps: bind_unpacked_struct_arg
 continued before the write-back registration, and the body's o.a = v write was dropped
 (the MemberAccess write arm never checked local_stack for dotted keys). Fix: the binding
 now returns per-member (local_key, caller_lvalue) pairs that both call sites write back
 after the body runs; plus a local-frame fallback for body writes.

aionhw:
Reviewed against current `main`. **The fix is wanted** — `tests/class_unpacked_struct_property.rs` from this PR fails on `main` today, so the whole-value unpacked-struct class-property path is genuinely broken. (`struct_output_formal` and `type_param_struct_formal` already pass, so those are welcome extra coverage.)

**Needs a rebase before it can merge.** `src/compiler/simulator.rs` conflicts with `main`, but only textually: both this PR and a recent `main` commit add a new arm immediately after the same `return changed;` in the `MemberAccess` lvalue path of `assign_value`. The two conditions are disjoint —

* `main`'s arm: an element of a queue / dynamic / associative collection whose element type is an unpacked struct (`u.q[0].field = v`), gated on the container tables;
* this PR's arm: a struct FORMAL bound member-wise, gated on the dotted key being present in `local_stack.last()`.

so keeping **both**, in either order, is correct. No semantic clash — just re-apply on top of `main` and the conflict disappears.

Once rebased I'll re-run the full suite and merge.

Thomasb81:
Rebased onto current main (head is now 52328fb, force-pushed). The conflict flagged is resolved — in assign_value's
 MemberAccess lvalue path I kept both arms: the queue/dynamic/assoc collection-elem struct-field write (from main) and the member-wise
 struct-formal write-back (this PR). They're disjoint and each self-contained.

 The three self-tests added by this PR all pass:
 - class_unpacked_struct_property — read/write/compare of unpacked struct class properties
 - struct_output_formal — output/inout/ref struct formals written back member-wise
 - type_param_struct_formal — type-parameter struct formals bound member-wise

 One note for when you re-run the full suite: randomize_fills_fields_wider_than_64_bits currently fails with a subtract-overflow in the constraint-bounds path — but this is pre-existing on main (296df69), not introduced here. I confirmed it fails identically on a clean checkout of main before applying these commits; this PR's changes don't touch the randomize code.

Thomasb81:
**Rebase complete — head is now `795c8e4` on top of `main` (`17f2fc4`), MERGEABLE.**

While re-applying commit 3 (output/inout/ref struct formals) I found and fixed a gap that the rebase surfaced: main's recent `5f2e35f` changed how subroutine-local unpacked structs are stored. The `inout`/`ref` free-function sub-cases (2 of 5) regressed because the copy-back write landed in a phantom whole-struct slot while the actual value lived in standalone dotted-key leaves. Fixed by writing to the existing dotted leaf directly when one is present. All 5 sub-cases in `struct_output_formal` now pass, byte-for-byte identical to the reference simulator.

The three self-tests all pass:
- `class_unpacked_struct_property`
- `type_param_struct_formal`
- `struct_output_formal` (all 5 sub-cases)

**On the CI failures — both are pre-existing on `main`'s own CI, not introduced here.** Upstream main `17f2fc4` run [#31040830437](https://github.com/aionhw/xezim/actions/runs/31040830437) fails identically:
- `uvm_config_db_tests::test_config_db_*` (×3) — `could not read 1800.2 uvm_pkg.sv: NotFound`. The test reads the UVM library from a sibling dir (`../1800.2-2020.3.1/`) that isn't checked out in CI. (These pass locally where the library is present.)
- `randomize_fills_fields_wider_than_64_bits` — the 128-bit subtract-overflow in the constraint-bounds path, already noted on the previous base.

No other suite changes; this PR touches only `simulator.rs` and adds the three test files.

Thomasb81:
uvm_config_db_tests::test_config_db_* issue is fixed with #87

Thomasb81:
Rebased onto current `main` (`1f7fae5`). The branch was 13 behind and had fallen CONFLICTING.

**Conflict:** one block in `simulator.rs`, in the class-property read path. Two orthogonal features landed on the same spot — this PR's whole-struct-class-property read assembly (`read_whole_struct_class_prop`) and main's recent `shadowed_prop_key` (§8.10 property shadowing, from `1f7fae5`). Resolved by composing both: resolve the shadowed key first, then attempt struct assembly on that key, then fall through to the property lookup. Neither feature is lost.

**Complementary to main's `402a04d\**:"`unpacked-struct class properties resolve at every depth and per instance" fixes member resolution / per-instance leaks; this PR adds the *whole-value* view (read/assign/compare assembly, type-parameter struct formals, `output`/`inout`/`ref` struct formals). All three of this PR's tests still fail on clean `main` and pass with it.

**Other:**
- Test files relocated into `tests/classes/` per the #83 group structure (no new link units).
- All three tests byte-for-byte identical to a reference simulator (`pass_count=3`).
- The four classes-group failures on CI (`module_scope_derived_constraints` ×2, `randomize_member_subset` ×2) reproduce on clean `main` — not introduced here.

PR #87, opened 2026-08-05
Retire PURE_SV_LRM=0 shims, run UVM natively (replaces #79)

Thomasb81:
**Replaces #79.** This is the same work re-applied onto current `main` (`17f2fc4`) as a single clean commit, in response to the review on #79.

## Why a new branch

#79's branch had fallen 40+ commits behind `main`, and GitHub reported it `CONFLICTING/DIRTY`. More importantly, the 7 regressions found in review turned out to be **stale-base interaction artifacts**, not consequences of the changes themselves — I confirmed this by checking out #79's branch directly: all 14 `udp_primitives` tests pass there. The regressions only appeared on a 3-way apply against current `main`, because they were interactions with `main`'s state at that time.

Re-applying onto current `main` clears the conflicts and **eliminates the regressions entirely** — the full self-test suite (1048 tests) passes at the clean-`main` baseline.

## Verification

```
cargo test → 1048 passed, 1 failed
```

The single failure is `randomize_fills_fields_wider_than_64_bits`, which is **pre-existing on `main`** (`17f2fc4`) — unrelated to this PR.

The 7 tests that regressed in #79's review all pass here:

| test (was failing) | now |
|---|---|
| `udp_primitives` (was 9/14) | **14/14** |
| `library_flags` (primitive_verbose, v_file) | **6/6** |
| `nonvirtual_dispatch_fscanf_process` | **4/4** |
| `format_lrm_compliance` | **3/3** |

I also verified the rebase preserved all of #79's content: every deleted function (20), every added code line (83), and every straggler-cleanup edit survived the squash-merge + rebase.

## What this does

Accellera UVM (IEEE 1800.2-2020.3.1) now runs via standard SystemVerilog semantics — native class inheritance, method dispatch, process control, associative arrays and DPI bindings — instead of simulator-level shims.

**Removes (~2220 lines net):**
- The `pure_sv_lrm` flag and its `PURE_SV_LRM` env-var wiring.
- Hardcoded UVM method dispatch/interceptors: `run_test` routing, `config_db` virtual-interface interception, `factory::create` shims, `uvm_report` / `get_is_active` / `sprint` backstops, `uvm_glob_match`.
- Simulator-side objection/phase tracking: `uvm_obj_count`, `uvm_obj_raised`, `uvm_phase_drain`, `uvm_pending_end`, `genuine_uvm_finished`, and the `maybe_end_run_phase` / `run_uvm_cleanup_phases` / `end_run_phase` / `handle_uvm_objection` / `bridge_objection_wait_for` backstops they fed.
- Legacy mock headers (`tests/uvm/uvm_mock.svh`) and the mock integration test (`tests/uvm_integration_tests.rs`), replaced by pure-SV self-checking tests (`pure_sv_phase_objection`, `procedural_loop_stall`).

## The two design-semantic decisions you flagged

These were previously gated behind `pure_sv_lrm`; the gate is now gone, so they apply to every design. Both are correct per the LRM and validated by the full suite:

1. **`mailbox.get` counts as blocking.** A `mailbox.get` in a forever loop is now routed through the scheduler so its wait suspends. This is LRM-correct: `mailbox.get` **is** a blocking statement (§15.5).
2. **Unknown bare `TypeReference` defaults to `null`.** A `TypeReference` that names neither a known width-typedef nor a known class defaults to `null` (0) instead of `x` — the class-handle alias idiom (`rsrc_q_t rq; if (rq==null) rq=new();`).

Happy to split these two into separate follow-up commits/PRs if you'd prefer each gated decision justified on its own; I kept them here because the suite confirms both are safe on current `main`.

Related to #73.


Thomasb81:
Rebased onto current `main` (`1f7fae5`). The branch was 13 behind.

**Conflicts (3, all in `simulator.rs`):**

1. `runtime_recv_class` — kept this PR's `eval_handle_expr(recv)` refactor. The helper is a superset of main's inline resolution (handles `this`/`super`/multi-segment/`MemberAccess`), so the behavior is preserved and generalized.
2. `expr_assoc_name` — kept main's side (`|| obj == "super"`), the newer §8.15 fix this PR didn't have.
3. method-call dispatch — combined both: kept main's new interface-subroutine dispatch block (§25.7/§7.10), then applied this PR's removal of the `uses_real_uvm()` interception line.

**Validation — no regressions:**
- Self-tests: the 8 failures (`module_scope_derived_constraints` ×2, `randomize_member_subset` ×2, `interface_array_element_ports` ×3, `randomize_fills_fields_wider_than_64_bits` ×1) all reproduce identically on clean `main`.
- UVM native execution: byte-for-byte identical to the pre-rebase commit on a Mantis test (4160 produces the same `MISCMP .a lhs='hb27281e9 rhs='h0` both ways). The rebase changed no UVM behavior.

PR #90, opened 2026-08-06
ci: parallelize the test suite with a build-once, shard-run matrix

opensource-elearning:
# ci: parallelize the test suite with a build-once, shard-run matrix

## What this changes

The single sequential CI job is split into a **build** job and an 18-cell
**test matrix** so all test groups run concurrently instead of back-to-back
on one runner.

- **build** — compiles every test binary once (both feature sets) and caches
  `target/` together with the `xezim-core` path dependency (mtimes preserved)
  so cargo's fingerprinting does not rebuild it downstream.
- **test** — one job per `(group, features)` cell restores the shared build
  and runs only its own binary with `--test-threads=8`.
- **coverage is unchanged**: lib unit tests, doctests, and all 8 integration
  binaries run — the identical set the sequential job covered.

## Why

Linking dominates CI wall-clock — the suite itself *runs* in seconds once the
binaries exist. The old job linked every binary **twice** (once per feature
set) and ran every group **sequentially**, paying the full link price four
times. Building once and sharding the run lets each group's run overlap on
its own worker.

## Measured gains

Local measurements with the same binaries on the same machine:

| | Sequential (old CI shape) | Parallel (this PR) |
|---|---|---|
| **All 8 groups, run phase** | ~110 s | ~35 s wall-clock (8 workers) |
| **Link cost paid** | 4× (2 builds × 2 feature sets) | 1× |
| **Warm-cache test phase** | ~110 s back-to-back | ~max(group) ≈ 34 s |
| **`lib` + doctests** | after every other group | concurrent with everything |

### Verification (identical results)

| Target | Result | Time |
|---|---|---|
| classes | 174 pass · 3 env-missing · 2 ignored | ~2 s |
| collections | 93 pass · 1 ignored | ~0.6 s |
| gates | 59 pass | ~0.3 s |
| hierarchy | 93 pass | ~0.4 s |
| misc | 585 pass · 1 known · 9 ignored | ~30 s |
| scheduling | 259 pass | ~34 s |
| strings | 106 pass | ~1.5 s |
| types | 249 pass | ~31 s |
| lib unit | 31 pass | 0.02 s |
| doctests | 1 ignored | 0 s |

Same tests, same names, same assertions — only the link/run scheduling
changed. No test was removed, renamed, or skipped.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,13 @@ env:
   RUST_BACKTRACE: 1
 
 jobs:
-  test:
+  # Build every test binary ONCE, then let the `test` matrix jobs below
+  # restore this `target/` AND `xezim-core/` and only *run* the group they
+  # own. Linking is the dominant cost (the suite itself runs in seconds once
+  # built), so compiling once instead of once-per-group keeps cold-cache CI
+  # from paying the link price eighteen times. The `xezim-core/` cache must
+  # preserve mtimes so cargo's fingerprint does not rebuild path dependencies.
+  build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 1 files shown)


aionhw:
**Reviewed — closing for now.** The build-once / shard-run shape is the right idea, but the matrix silently drops test coverage as written, and that needs fixing before it lands.

### The matrix omits the `perf` group

```yaml
test-group: [classes, collections, gates, hierarchy, misc, scheduling, strings, types, lib]
```

`main` currently has **nine** integration test binaries:

```
classes collections gates hierarchy misc perf scheduling strings types
```

`perf` is not in the list, so `tests/perf.rs` (2 tests) would stop running in CI entirely. The PR comment describes the old job as covering "lib + doc + all 8 integration binaries" — that was accurate when written, but `perf` makes nine, and the previous `cargo test --no-fail-fast` picked it up automatically.

### The underlying issue is the hardcoded list

Enumerating groups by hand means CI coverage silently shrinks whenever someone adds a test binary, with nothing to catch it. That is already about to bite twice more: #93 and #94 each add new top-level `tests/*.rs` files, and neither would appear in this matrix.

Two ways out:
- derive the list (e.g. a small `prepare` job that globs `tests/*.rs` and emits a matrix via `fromJSON`), or
- keep the hardcoded list but add a guard step that fails when `ls tests/*.rs` does not match it.

Either keeps the parallelism win without trading away coverage.

### Worth re-running

Note this was authored against a red `main` — both `msrv` and `test` were failing on `f0f842f`, so the 4 failing shards here could not be distinguished from the broken baseline. `main` is green now, so a rebase would give this real signal.

Happy to see this come back with the group list fixed — the wall-clock benefit is worth having.

PR #92, opened 2026-08-06
fix: avoid i128 underflow for unsigned rand domains at width 127

opensource-elearning:
# fix: avoid i128 underflow for unsigned rand domains at width 127

## Problem

An unsigned rand field of width >= 128 clamps its domain to
`[0, 2^127 - 1]` per §18.4. The upper bound was computed as
`(1i128 << 127) - 1`, which underflows in debug builds: `1i128 << 127`
sets the sign bit (i128::MIN), so `- 1` wraps to i128::MAX - 1 but
panics as an arithmetic underflow. This caused the native test
`randomize_fills_fields_wider_than_64_bits` to fail on the CI.

## Fix

At the three domain-computation sites, use `i128::MAX` directly when
`width == 127`. The bound is mathematically correct (`2^127 - 1 ==
i128::MAX`) and avoids the underflow.

## Verification

The previously failing test `randomize_fills_fields_wider_than_64_bits`
now passes. All other tests unchanged — same failure set as upstream
minus the one fixed test.

```diff
- test result: FAILED. 585 passed; 1 failed; 9 ignored
+ test result: ok. 586 passed; 0 failed; 9 ignored
```

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -56419,7 +56419,10 @@ impl Simulator {
                                             (1i128 << ww.saturating_sub(1)) - 1,
                                         )
                                     } else {
-                                        (0, (1i128 << ww) - 1)
+                                        // §18.4: unsigned domain [0, 2^ww - 1];
+                                        // for ww == 127 that bound is i128::MAX,
+                                        // and the shift form underflows in debug.
+                                        (0, if ww == 127 { i128::MAX } else { (1i128 << ww) - 1 })
                                     };
                                     cvar_names.push(name.clone());
                                     cvar_dom.push((lo, hi));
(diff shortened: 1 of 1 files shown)


opensource-elearning:
Keeping this as the canonical PR for the width-127 rand fix. It supersedes #91 (duplicate, now closed): identical change, but based on the newer `main`.

Heads-up on CI: the `test` job fails on the pre-existing `uvm_config_db_tests` panic (tests/classes/uvm_config_db_tests.rs:19) which is also failing on `main` — unrelated to this change. `msrv` passes, and the native rand regression suite that this PR targets now reports `586 passed; 0 failed`.

aionhw:
Superseded — this fix is now on `main` as commit `66a6f2c`, which reached the same conclusion your patch did: the unsigned domain bound at width 127 is `i128::MAX`, because `1i128 << 127` is `i128::MIN` and the `- 1` overflows in debug builds.

Credit for the diagnosis is yours — I hit the same panic while unblocking CI and landed it directly rather than noticing your PR already had it. Apologies for the duplicated effort.

Two notes from reviewing your version:
- It was **correct and complete**. The two other `(1i128 << w) - 1` sites (`solve_affine_eq` and the class-property domain) both bail at `width > 64`, so they cannot reach the boundary — you were right to leave them alone.
- The failing test on `main` (`misc::child_decl_init_and_wide_rand::randomize_fills_fields_wider_than_64_bits`) was already covering this; it only reproduces in **debug**, since release wraps silently. Anyone running `cargo test --release` would never have seen it.

Closing as merged-by-other-means, not rejected.

Issue #18, opened 2026-07-10
User-Defined Data Types as Parameters


Ganesh-TNL:
Sample test case on EDA Playground:

https://www.edaplayground.com/x/bUKH

Test case passes only with Aldec Riviera Pro. Hopefully, xezim can match that.

Other commercial simulators fail on elaboration, as the new function being called inside the parameterized class can't be supported for non-class data types. Ideally, that should error out only at run-time in case of user-error in calling unsupported functions of the parameterized class - I believe Riviera Pro would do that.

Currently, xezim 0.8.3 seems to be able to handle only built-in data types. Does this require any specific simulator option to work? I am curious because UVM 1.2 seems to be supported based on many passing tests in the repo. UVM backend is SystemVerilog and it heavily uses classes parameterized by type.

<details>

<summary>Output log from xezim 0.8.3</summary>

```
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim type-parameter-compliance.sv
=== xezim 0.8.3 ===
Max time: 100000
------------------------------
[PHASE] compilation: 5.1ms
[EVENT-EDGE] measure (timestamp): 0 edge blocks, 0 gateable, 0 gateable-with-EMPTY-data-reads, avg data-reads/gateable=0.00
[EVENT-EDGE] SKIP mode ON (sim_time >= 0) — skipping gateable flop fires with no data-input change
================================================================
STARTING IEEE STD 1800-2017 SEC 6.20.3 TYPE PARAMETER COMPLIANCE
================================================================
[TEST] Verifying built-in types as parameters...
[TEST] Verifying user-defined packed struct as parameter...
[ERROR] User-defined struct type parameter failed layout retention.
[TEST] Verifying user-defined unpacked array as parameter...
[ERROR] User-defined array type parameter mapping failed.
[TEST] Verifying custom CLASS handles as type parameters...
[ERROR] Properties inside the type-parameterized class are corrupted.
================================================================
TEST FAILED: 3 type parameter errors detected.
================================================================
[PROF] settle=0.0ms edges=0.0ms nba=0.0ms process=2.6ms snap=0.0ms sched=0.0ms
[PROF] edge_waiters=0.0ms edge_cg=0.0ms waiter_iters=0
[PROF] clocks_only_detect=0 nba_elided=0 (iters with clocks-only scan / NBAs dropped at eval because value matched signal_table)
[...]

Issue #27, opened 2026-07-13
Upward Name Referencing from Bound Modules


rharikrishna25:
The following SystemVerilog test case fails with xezim v0.9.0:

<details>

<summary>SV Code with Upward Name Referencing via Bound Modules ( sv_bind_upward_name_referencing_tb.sv )</summary>

```systemverilog
`ifndef SVTEST_DEFS_SVH
`define SVTEST_DEFS_SVH

`define SVTEST_INIT \
  int failures = 0;

`define SVTEST_CHECK(expr, msg) \
  if (!(expr)) begin \
    failures++; \
    $display("FAIL: %s", msg); \
  end

`define SVTEST_PASSFAIL \
  if (failures == 0) begin \
    $display("TEST_PASS"); \
  end else begin \
    $display("TEST_FAIL count=%0d", failures); \
    $fatal(1); \
  end

`endif


// 1. Package-level definition (Upward Reference Flavor A)
package my_pkg;
  int pkg_secret = 99;
endpackage

module dut_top;
  // 2. Lexical/Module-level definition (Upward Reference Flavor B)
  int top_secret = 42;

  // Intermediate hierarchy
  sub_block u_sub_block();
endmodule

module sub_block;
  // 3. Nested-level definition (Upward Reference Flavor C)
  int sub_secret = 7;

// Target instance for the bind
  target_core u_target_core();
endmodule

module target_core;
  // This block is empty in the original design.
  // It will receive the bound monitor module.
endmodule

module bind_monitor;

  initial begin
    // Instantiate macro to create 'failures' variable
    `SVTEST_INIT

    $display("--- Starting Bind & Upward Referencing Verification ---");

    // Flavor 1: Upward Hierarchical Reference using explicit parent instance names
    // SystemVerilog searches upwards through the instantiation tree for these names.
    `SVTEST_CHECK(dut_top.top_secret == 42, "Failed to read upward reference: dut_top.top_secret")

    `SVTEST_CHECK(sub_block.sub_secret == 7, "Failed to read upward reference: sub_block.sub_secret")

    // Flavor 2: Lexical Upward Reference via the binding context
    // Because the bind happens inside the module namespace, some simulators allow 
    // direct lookups of scopes visible to the parent context.
	// Doesn't work on ANY commercial simulator
    // `SVTEST_CHECK(sub_secret == 7, "Failed to read lexical upward reference: sub_secret")

    // Flavor 3: Package Scoping / Scope Resolution (Global Upward Reference)
[...]

PR #97, opened 2026-08-07
sim: spread struct-typed nettype resolver results into member nets (IEEE1800-2017 §6.6.7)

opensource-elearning:
# sim: spread struct-typed nettype resolver results into member nets (IEEE1800-2017 §6.6.7)

Companion to the xezim-core fix **"fix: dispatch user-defined nettype
resolver functions (IEEE1800-2017 §6.6.7)"** —
[aionhw/xezim-core#18](https://github.com/aionhw/xezim-core/pull/18).
Merge the core change first: xezim CI builds against xezim-core at `main`, so
this regression test only passes once the paired core PR lands.

## Problem

When a nettype's element type is an **unpacked struct**, the net is stored
member-wise — one leaf signal per member (§7.2). The resolver returns a
packed struct value (built from the driver queue), and the continuous assign
wrote that packed value into the net's container signal, which no member
read ever touches. Every member leaf stayed x.

Additionally, `spread_into_unpacked_struct` reinterpreted a real member's
raw IEEE-754 bits as an integer (`mv.to_u64() as f64`), so e.g. `4.0`
(`0x4010000000000000`) became `4613937818241073152.0`.

## Fix

- In the continuous-assign arm, when the LHS signal has a struct-typed
  nettype (resolved via `signal_type_names` + `unpacked_struct_of`), spread
  the packed RHS across the member leaves with
  `spread_into_unpacked_struct` instead of the generic scalar store.
- In `spread_into_unpacked_struct`, rebuild a real member from its copied
  bits with `f64::from_bits`, not `to_u64() as f64`.
- Adds `tests/types/nettype_resolver_dispatch.rs`, which also proves a
  **sum** resolver is dispatched (0x12+0x14 = 0x26) rather than OR-folded
  (0x16).

## Verification

- Paired repro above: members read back `x`/`0` before, correct after.
- The bundled compliance suite advanced 38 and 40 → `TEST_PASS`.
- `cargo test --no-fail-fast` and `cargo test --features jit --no-fail-fast`:
  `tests/types` 253/253 green.

## Checklist

- [x] Failing behavior captured before the fix
- [x] Passes with the fix
- [x] LRM reference: IEEE 1800-2017 §6.6.7, §7.2
- [x] Regression test follows the `tests/types` convention (group root mod line added)
- [x] No unrelated changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,8 +10,8 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 ahash = { version = "0.8", features = ["serde"] }
 rand = "0.8"
-xezim-core = { path = "../xezim-core" }
-sv-parser = { path = "../xezim-core/xezim-parser" }
+xezim-core = { path = "/home/temp/wt-core18-rebase" }
+sv-parser = { path = "/home/temp/wt-core18-rebase/xezim-parser" }
 libloading = "0.8"
 libffi = "3"
 libc = "0.2"
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -13,7 +13,8 @@ use crate::ast::decl::{
 };
 use crate::ast::expr::*;
 use crate::ast::stmt::*;
-use crate::ast::types::{DataType, IntegerAtomType, PortDirection};
+use crate::ast::types::{DataType, IntegerAtomType, PortDirection, TypeName};
+use crate::ast::{Identifier, Span};
 #[allow(unused_imports)]
 use crate::{log_eprintln as eprintln, log_println as println};
 use fst_writer::{
@@ -32189,23 +32190,44 @@ impl Simulator {
                         }
                     }
                 } else if let Some(id) = lhs_id {
-                    let width = self.signal_widths[id];
-                    let mut resized = if self.signal_real[id] {
-                        if val.is_real {
-                            val.clone()
-                        } else {
-                            Value::from_f64(val.to_f64())
+                    // IEEE 1800-2023 §6.6.7: struct-typed nettype with resolver.
+                    // The resolver returns a packed struct value (from the queue
+                    // argument). The LHS is an unpacked struct stored member-wise.
+                    // Spread the packed result into the unpacked leaf signals.
+                    let type_name = self.signal_type_names.get(&id).cloned();
+                    let mut spread_done = false;
+                    if let Some(ref tn) = type_name {
+                        if let Some(su) = self.unpacked_struct_of(&DataType::TypeReference {
+                            name: TypeName { scope: None, name: Identifier { name: tn.clone(), span: Span::dummy() }, span: Span::dummy() },
+                            dimensions: Vec::new(),
+                            type_args: Vec::new(),
+                            span: Span::dummy(),
+                        }) {
+                            let name_str: String = self.id_to_name.get(id).map(|s| s.as_ref().to_string()).unwrap_or_default();
+                            if self.spread_into_unpacked_struct(&name_str, &su, &val) {
+                                spread_done = true;
+                            }
                         }
-                    } else if val.is_real {
-                        Self::real_to_int(val.to_f64(), width)
-                    } else {
-                        val.resize(width)
-                    };
-                    resized.is_signed = self.signal_signed[id];
-                    if self.signal_table[id] != resized {
-                        self.mark_dirty_id(id);
-                        write_sig!(self, id, resized);
-                        self.table_modified = true;
+                    }
+                    if !spread_done {
+                        let width = self.signal_widths[id];
+                        let mut resized = if self.signal_real[id] {
+                            if val.is_real {
+                                val.clone()
+                            } else {
+                                Value::from_f64(val.to_f64())
+                            }
+                        } else if val.is_real {
+                            Self::real_to_int(val.to_f64(), width)
+                        } else {
+                            val.resize(width)
+                        };
+                        resized.is_signed = self.signal_signed[id];
+                        if self.signal_table[id] != resized {
+                            self.mark_dirty_id(id);
+                            write_sig!(self, id, resized);
+                            self.table_modified = true;
+                        }
                     }
                 } else {
                     self.assign_value(lhs, &val);
@@ -70999,8 +71021,12 @@ impl Simulator {
                 mv.set_bit(i as usize, v.get_bit((off + i) as usize));
             }
             if is_real {
-                let f = mv.to_u64().unwrap_or(0) as f64;
-                self.write_leaf_by_name(&leaf, Value::from_f64(f));
+                // mv holds the raw IEEE-754 BITS copied from the packed value.
+                // Rebuild the double from those bits — `to_u64() as f64` would
+                // reinterpret e.g. 0x4010000000000000 (4.0) as the integer
+                // 4613937818241073152.0, corrupting the leaf.
+                let bits = mv.to_u64().unwrap_or(0);
+                self.write_leaf_by_name(&leaf, Value::from_f64(f64::from_bits(bits)));
             } else {
                 self.write_leaf_by_name(&leaf, mv);
             }
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -2743,6 +2743,14 @@ pub struct Simulator {
     /// dropped end took a `{base, "TAIL"}` suffix with it). These ids skip the
     /// width fit and keep whatever length they are given.
     signal_is_string: Vec<bool>,
+    /// Signal ids whose declared type resolves to an unpacked struct stored
+    /// member-wise (§7.2), i.e. the continuous-assign fallback must SPREAD a
+    /// packed RHS across the member leaves instead of storing it whole in the
+    /// container signal. Fixed at elaboration time, so it is precomputed once
+    /// in `Simulator::new`; the per-evaluation gate is a single indexed read
+    /// (`spread_into_unpacked_struct` + the typedef chain behind it only run
+    /// for signals actually in this set).
+    signal_spread_struct: Vec<bool>,
     /// `--warn-x` state. `warn_x` is a plain field (not the atomic) so the
     /// write hot path costs one bool load; everything else is touched only
     /// when a report actually fires.
@@ -6132,6 +6140,7 @@ impl Simulator {
             signal_two_state,
             signal_gate_driven,
             signal_is_string,
+            signal_spread_struct: vec![false; num_signals],
             warn_x: warn_x_enabled(),
             warn_x_seen: std::cell::RefCell::new(HashSet::default()),
             warn_x_pending: std::cell::RefCell::new(Vec::new()),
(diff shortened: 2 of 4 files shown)


opensource-elearning:
CI note: the only new failure introduced here is `nettype_resolver_dispatch::nettype_resolver_is_called_not_or_folded`. xezim CI builds against xezim-core at `main` (the workflow clones aionhw/xezim-core), so this regression test needs the paired xezim-core change (PR #18, resolver dispatch) to land first.

The other failing tests — `uvm_config_db_tests` (x3) and `child_decl_init_and_wide_rand::randomize_fills_fields_wider_than_64_bits` — are pre-existing on `main`: main's own CI is currently red with the same four. No new regression from this PR beyond the cross-repo pairing.

aionhw:
**Needs an update, but the direction is right.** Two things, plus a correction to something I got wrong in an earlier pass.

### 1. This is blocked on xezim-core#18

Verified by building them apart. With this PR merged onto current `main` and xezim-core#18 *not* merged, its own test fails:

```
nettype_resolver_dispatch::nettype_resolver_is_called_not_or_folded ... FAILED
assertion `left == right` failed: multi-driver nettype resolver sum
  left: 22   right: 38
```

`22` is `0x16` (the old hardcoded OR-fold), `38` is `0x26` (the resolver's sum) — exactly the distinction the test was written to catch. So the two halves have to land together, core first.

Worth knowing that CI cannot catch this: `.github/workflows/*.yml` clones xezim-core from **upstream main**, never the paired branch, so the two sides of a split change are never compiled together by any automated check. Same situation applies to #89/core#17 and #98/core#19.

Note also that xezim-core#18 currently drops every driver when a nettype has no resolver (the `kept.push` only runs inside the two nested `if let`s, and the grouped drivers were already removed from `kept` above). That should be fixed before this lands on top of it.

### 2. Correction: the perf concern is narrower than I first said

I previously described the new block as sitting in the settle hot path and costing every design an allocation per continuous assign. That was wrong and I want to correct it publicly rather than leave it standing.

The block is in `eval_ast_comb_entry`, which the settle loop reaches only for `CombItem::ContAssign` — the **fallback for assigns the bytecode compiler could not compile** (`simulator.rs:17844`, `17904`). It is not the normal path. I measured with a 300-signal typedef'd-struct design over 2.4M entry evaluations:

```
[PROF] settle_dc=2400600  settle_ca=0  settle_ab=0
```

Zero evaluations reached the modified code; they all fused to DirectCopy. So this is **not** a broad regression, and I apologise for implying it was.

### 3. The cost that does remain, when the path is taken

Per evaluation the block does `signal_type_names.get(&id).cloned()`, then a second `tn.clone()` into a freshly built `DataType::TypeReference`, then `unpacked_struct_of` → `resolve_dt`:

```rust
let mut cur = dt.clone();
for _ in 0..16 {
    if let Some(next) = self.module.typedef_types.get(&name.name.name) {
        cur = next.clone();   // deep clone of the whole resolved DataType
```

For a struct typedef that deep-clones the entire `StructUnionType` — every member, declarator and dimension expression — potentially several times down the typedef chain, and `unpacked_struct_of` then returns it by value again. That is unbounded AST cloning to answer a question whose answer is fixed at elaboration time.

Cheap fix: resolve the set of resolver-backed nettype signal ids once when the comb entries are built, store it as a `HashSet<usize>` (or a bitset parallel to `signal_widths`), and gate the whole block on membership. Common case becomes one integer lookup, behaviour unchanged.

### 4. Unrelated good fix worth splitting out

The `f64::from_bits` hunk is a real, independent bug fix — `mv.to_u64().unwrap_or(0) as f64` was reinterpreting the raw IEEE-754 bits as an integer, so a `real` struct member read back as e.g. `4613937818241073152.0` instead of `4.0`. That has nothing to do with nettypes and would land on its own immediately, without waiting on core#18. Might be worth pulling into a separate PR so the fix is not held up.

---

For reference, `main` is now green (`msrv` + `test`) after the MSRV and fixture fixes, so a rebase will give this a baseline where a red check actually means something.

opensource-elearning:
Addressed the review feedback and rebased onto green `main` (both `test` and `msrv` now pass on `main`).

### 1. Cross-repo pairing with xezim-core#18

The paired core PR is updated and rebased too. It also now fixes the no-resolver driver-drop bug the review called out: a nettype without a resolver previously dropped every driver (the grouped drivers were already removed from `kept` before the resolver lookup), leaving the net permanently `x`. The no-resolver case falls back to the legacy OR-combined fold so every driver is kept. Verified with a standalone repro: `PASS: nettype w/o resolver OR-folds -> 0x16`.

### 2. Per-evaluation struct-type resolution in the spread fallback

The review's suggestion is implemented: instead of re-resolving the declared-type chain per evaluation (`signal_type_names.get` clone + fresh `DataType::TypeReference` + `unpacked_struct_of` → `resolve_dt`), the set of signal ids that need the member-wise spread is precomputed once in `Simulator::new` into a bitset parallel to `signal_widths`. The continuous-assign fallback now gates on a single indexed read; the typedef-chain resolution only runs for signals actually in the set. Behaviour unchanged.

### 3. `f64::from_bits` in the spread

Kept in this PR rather than split out: it is the member-wise spread's real-member path, and both sides of the change land together now (core first), so it is not held up by the pairing.

### Verification (local, same commands CI runs)

- `cargo test` — all targets green on the rebased branch.
- `cargo test --features jit --test types nettype_resolver_dispatch` — green.
- Compliance tests 38 (scalar resolver dispatch) and 40 (struct nettype with `real` member, field-by-field resolver) run to `TEST_PASS`.

opensource-elearning:
Rebased onto current `main` (02ff84e). The branch was stale (based on 9d7af43, one commit behind main's chained-selects fix), so the old diff showed stale deletions of main's newer tests (`packed_3d_chained_select`, `bits_of_type_operands`, `void_cast_queue_ops`) — those are gone now, leaving a minimal subject:

- `src/compiler/simulator.rs`: struct-typed nettype spread logic (gate `signal_spread_struct` + `spread_into_unpacked_struct` with `f64::from_bits` for real members).
- `tests/types.rs`: entry point for the new `nettype_resolver_dispatch` test.
- `tests/types/nettype_resolver_dispatch.rs`: new test proving resolver dispatch for scalar nettypes (sum resolver: `0x12 + 0x14 = 0x26` vs OR-fold `0x16`).

**Cross-repo dependency:** this PR's test passes against the paired core PR #18 (`fix: dispatch user-defined nettype resolver functions`), which must merge first — xezim CI builds against `xezim-core@main`, so the test will be green once core #18 lands. Verified locally:

- core #18 rebased onto current core `main` (7a4d26c).
- `cargo test` and `cargo test --features jit` all green against the rebased core #18.

When core #18 merges, xezim CI will pick it up and this PR will go green.

opensource-elearning:
Cross-referencing the core-side discussion on the repo-split friction: https://github.com/aionhw/xezim-core/pull/18#issuecomment-5225358903

This PR (#97) is a concrete example — the test depends on core#18 (resolver dispatch), but xezim CI builds against `xezim-core@main`, so the test fails on CI until core#18 merges. We validated the pair locally with two worktrees + Cargo overrides, but that's manual work every contributor pays.

A submodule (or monorepo) would make "core + xezim at matching commits" the default CI state, eliminating this class of cross-repo coordination overhead entirely.

opensource-elearning:
Cross-referencing the core-side discussion on the repo-split friction: https://github.com/aionhw/xezim-core/pull/18#issuecomment-5225370090

This PR (#97) is a concrete example — the test depends on core#18 (resolver dispatch), but xezim CI builds against `xezim-core@main`, so the test fails on CI until core#18 merges. We validated the pair locally with two worktrees + Cargo overrides, but that's manual work every contributor pays.

**Since `xezim-core` has no external consumers** (it's an implementation detail of the xezim simulator), a **monorepo** is the better long-term fit vs. submodule: single `cargo test` validates the whole stack, one PR = complete change, `git bisect` works across the boundary, and the `core/` crate can still be published to crates.io from the monorepo if release independence is desired. This eliminates the entire class of cross-repo coordination overhead.

aionhw:
**Request changes.**

### Blocking

1. **`Cargo.toml:11-12` repoints the core crates at an absolute machine-local path**: `xezim-core = { path = "/home/temp/wt-core18-rebase" }` and `sv-parser = { path = "/home/temp/wt-core18-rebase/xezim-parser" }`, replacing `../xezim-core`. That directory doesn't exist outside the author's machine, so this breaks the build for every other developer and for CI. It also means the PR's stated verification was produced against an unreproducible tree.

2. **The behavior being tested isn't in this diff** — it's in that worktree. §6.6.7 resolver *dispatch* is an elaborator concern; in-repo `xezim-core/src/elaborate.rs` still collapses multiple nettype drivers with a hardcoded `BitOr` fold. The new test asserts `m_out == 0x26` (a sum); against `aionhw/xezim-core` the same source produces `0x16`.

3. **The new spread branch is dead code against the in-repo core.** The gate needs the nettype name in `module.typedef_types`, but the nettype pre-pass (`xezim-core/src/elaborate.rs:3707`) inserts it only into `elab.typedefs` (a name→width map). So `unpacked_struct_of` returns `None` and `signal_spread_struct[id]` stays `false` for every signal.

4. **Only one of two identical continuous-assign fallbacks is patched.** The hunk lands in `eval_ast_comb_entry`; the byte-identical fallback in `eval_ast_contassign_entries` (used by `settle_combinatorial_perlp`, i.e. `XEZIM_DISPATCHER=perlp`) still stores the packed value whole. A struct nettype would then resolve differently depending on the selected dispatcher.

5. **No test covers the claimed change.** `tests/types/nettype_resolver_dispatch.rs:11` declares `typedef logic [7:0] byte_t` — a packed scalar. There is no struct anywhere in the fixture, so member spreading, non-trivial layouts, nested structs and arrays of struct members are all uncovered — that machinery is the bulk of this diff.

6. **The `signal_spread_struct` scan is much broader than nettypes.** It loops over all `signal_type_names` and sets the flag for any type resolving to a member-wise unpacked struct. Every ordinary unpacked-struct-typed signal that is the target of a continuous assign now takes the spread path instead of the whole-value store — a regression surface well outside user-defined nettypes. If the intent is nettype-only, gate it on `nettype_resolvers`.

### Extracted separately

7. The `f64::from_bits` fix at the tail is a **genuine bug** and I've landed it on its own with regression tests — thank you for spotting it. Reproduced against the reference: a struct with a `real` member assigned whole (function return or assignment pattern) read `4.0` back as `4.6161896e18`. It's unrelated to §6.6.7 and affects every unpacked-struct spread, so it belonged in its own change.

### Non-blocking
8. `signal_spread_struct` is allocated twice — `vec![false; num_signals]` in the struct literal, then reallocated and overwritten later.
9. `unpacked_struct_leaves` bails at `depth > 8` but `spread_into_unpacked_struct` still returns `true`, so a deeper nest silently drops leaves *and* skips the container write.
10. The spread path never writes the container signal itself, so a consumer reading `signal_table[id]` directly (aggregate VCD/FST dump) sees stale bits; the old code wrote it.
11. MSB/LSB ordering is correct — `struct_leaf_layout` reverses the flat leaf list so the first declared member lands at the top, recursively, and is self-consistent with `pack_unpacked_struct`.

opensource-elearning:

aionhw:
Closing as superseded: §6.6.7 resolver dispatch landed via core #27 + xezim #120 (approved, pending pin bump). The member-wise struct spread this PR did in the simulator is handled at elaboration there (whole-struct CAs expand into per-member assigns, keeping dependency tracking intact), so no simulator-side spread is needed. Note for future PRs: this branch's Cargo.toml points xezim-core at a local absolute path (/home/temp/wt-core18-rebase), which can't merge — keep the stock ../xezim-core path dep and pair core changes as a core PR.

PR #98, opened 2026-08-07
sim: recover the tag of a tagged-union pattern literal (IEEE1800-2017 §7.3.1)

opensource-elearning:
# sim: recover the tag of a tagged-union pattern literal (IEEE1800-2017 §7.3.1)

Companion to the xezim-core fix **"fix: parse tagged-union pattern literals
(IEEE1800-2017 §7.3.1)"** —
[aionhw/xezim-core#19](https://github.com/aionhw/xezim-core/pull/19).
Merge the core change first: xezim CI builds against xezim-core at `main`, so
this regression test only passes once the paired core PR lands.

## Problem

With parsing fixed, `u = tagged '{b: 8'h5a};` stored the value correctly
but every later `u.b` read returned x. A pattern literal carries no member
name between `tagged` and `'{` — the active member IS the first named item
of the pattern. The assignment left `active_union_tag` empty, and
`tagged_union_member` (which gates reads on the tag) then returned the
default x-value even though the stored bits were right.

## Fix

When `ExprKind::Tagged` has an empty tag and its inner expression is an
assignment pattern, derive the tag from the pattern's first `Named` item
before recording it in `active_union_tag`. The value path is unchanged.
Adds `tests/types/tagged_union_pattern_literal.rs` (including a re-tagging
case that re-selects the member).

## Verification

- Repro above: `u.b` read `x` before, `8'h5a` after.
- The bundled compliance suite advanced 36 → `TEST_PASS`.
- `cargo test --no-fail-fast` and `cargo test --features jit --no-fail-fast`:
  `tests/types` 253/253 green.

## Checklist

- [x] Failing behavior captured before the fix
- [x] Passes with the fix
- [x] LRM reference: IEEE 1800-2017 §7.3.1
- [x] Regression test follows the `tests/types` convention (group root mod line added)
- [x] No unrelated changes

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -42340,8 +42340,26 @@ impl Simulator {
                 if let ExprKind::Tagged { tag, inner } = &rvalue.kind {
                     if let ExprKind::Ident(lh) = &lvalue.kind {
                         let lname = self.resolve_hier_name(lh);
-                        self.active_union_tag
-                            .insert(lname.clone(), tag.name.clone());
+                        // §7.3.2: the pattern literal `tagged '{ member: value }`
+                        // carries no member name between `tagged` and `'{` — the
+                        // tag IS the first named item of the pattern. Without it
+                        // `active_union_tag` stays empty and every later
+                        // `u.member` read (tagged_union_member) returns x.
+                        let tag_name = if tag.name.is_empty() {
+                            inner.as_ref().map_or(String::new(), |ie| {
+                                if let ExprKind::AssignmentPattern(items) = &ie.kind {
+                                    items.iter().find_map(|it| match it {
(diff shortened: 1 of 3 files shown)


opensource-elearning:
CI note: the only new failure introduced here is `tagged_union_pattern_literal::tagged_union_pattern_literal_selects_member`. xezim CI builds against xezim-core at `main` (the workflow clones aionhw/xezim-core), so this regression test needs the paired xezim-core change (PR #19, tagged-union pattern literal parsing) to land first.

The other failing tests — `uvm_config_db_tests` (x3) and `child_decl_init_and_wide_rand::randomize_fills_fields_wider_than_64_bits` — are pre-existing on `main`: main's own CI is currently red with the same four. No new regression from this PR beyond the cross-repo pairing.

Issue #21, opened 2026-07-11
timescale Handling


Ganesh-TNL:
Sample test case on EDA Playground:

https://www.edaplayground.com/x/Bi4J

The above test case is primarily for 1800-2017 compliance testing.

For real-world workloads involving different files and different timescales, the simulator's command-line options are the best option.

While the documentation of different options for commercial simulators is under NDA, there is a helpful post <a href="https://www.linkedin.com/pulse/why-i-prefer-use-vtimescale-over-timescale-xcelium-dimitrijevic-bgpae/">here</a> about how Cadence Xcelium handles this problem. I would like to propose xezim implementing a similar -timescale / -vtimescale option.

<details>

<summary>Output log from xezim 0.8.5 for the compliance test case</summary>

```
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim timescale.test.sv
[warn] `timescale 1ns/100ps` declares precision finer than 1ns; sim ticks are 1ns
[warn] `timescale 1ns/1ps` declares precision finer than 1ns; sim ticks are 1ns
=== xezim 0.8.5 ===
Max time: 100000
------------------------------
[warn] `timescale 1ns/100ps` declares precision finer than 1ns; sim ticks are 1ns
[warn] `timescale 1ns/1ps` declares precision finer than 1ns; sim ticks are 1ns
[PHASE] compilation: 5.3ms
[EVENT-EDGE] measure (timestamp): 0 edge blocks, 0 gateable, 0 gateable-with-EMPTY-data-reads, avg data-reads/gateable=0.00
[EVENT-EDGE] SKIP mode ON (sim_time >= 0) — skipping gateable flop fires with no data-input change
================================================================
STARTING SINGLE-FILE TIMESCALE COMPLIANCE TEST (SEC 22.7)
================================================================
[INFO] Printing compiled design timescales via $printtimescale:

--- [TEST 1] Baseline Module (1ns / 100ps) ---
[INFO] M1 Pulse 1 Time: %0.3f ns (Expected: 1.600)
[ERROR] Failed to round 1.55ns up to 1.6ns grid precision.
[INFO] M1 Pulse 2 Time: %0.3f ns (Expected: 1.600)
[ERROR] Failed to drop sub-precision 0.04ns increment.

--- [TEST 2] Overridden Module (10ns / 1ns) ---
[INFO] M2 Pulse Time: %0.3f units (Expected: 2.400)
[...]

PR #99, opened 2026-08-07
UVM data-path fixes: class/AA/$bits/string properties + regression tests

Thomasb81:
Consolidated set of UVM data-path fixes and their regression tests, presented as a minimal, reviewable 4-commit history on top of `main`.

### Commits

1. **`fix(simulator)`** — consolidate UVM data-path fixes
   One squashed runtime change containing the simulator.rs fixes previously scattered across the integration history:
   - class-property packed-vector `$size` / `$left` / `$high`
   - wide→narrow local truncation and `$bits(<builtin>)` const-eval
   - associative-array key narrowing to declared index type
   - class `string` properties in `string_signals`
   - whole queue / dynamic-array member copies
   - `foreach` over a `string`'s content
   - unsigned rand-domain i128 underflow at width 127
   - packed multi-dim element select inside a size cast
   - string-formal frame scoping
   - struct-typed collection element field access
   - clocking / member-access-chain dispatch
   
   No functional behavior changes beyond the listed fixes.

2. **`tests(classes)`** — class dynamic-array/dynarray/foreach regression tests
   Six focused regression tests covering dynamic-array element sizing/new/fixed-copy, `$bits`-on-class-member shadowing, `foreach` dynarray element construction, and randomize of object arrays.
   Also removes `tests/classes/uvm_config_db_tests.rs` (its register relied on the retired `uvm_obj_count`).

3. **`tests(collections)`** — assoc-index narrowing, class collections, queue/struct element fixes
   Regression tests for AA index_key_width tracking, multidim associative exists/num, class collection `$bits` and whole-pattern handle assignment, struct-typed collection element field access, whole queue/dynarray copies, and automatic indexed-queue growth.

4. **`tests(types,misc)`** — `$bits` builtin-width, `$size` on class member, string-formal scope
   Regression tests for const-eval of `$bits(...)` in packed dims, `$size`/`$left`/`$high` on class-member packed vectors, and string-formal name scoping across sibling calls.

### Notes
- Historically **1754 passed / 0 failed** on this exact tree.
- Depends on aionhw/xezim-core PR (ForeachTail.key_type / assoc_index_widths / assoc_index_props) for the remaining #99 testcases currently held back.


aionhw:
**Reviewed against current `main`.** Short version: this branch cannot be cherry-picked from as it stands, because its `xezim-core` half was never opened. I have taken the four testcases that stand alone; the rest need that core PR first.

### The branch no longer builds

Checking out this head and building against `xezim-core` main gives:

```
error[E0026]: variant `ForeachTail` does not have a field named `key_type`
   --> src/compiler/simulator.rs:26810
error[E0609]: no field `assoc_index_widths` on type `ElaboratedModule`
   --> src/compiler/simulator.rs:56297
error[E0609]: no field `assoc_index_props` on type `&ElaboratedClass`
   --> src/compiler/simulator.rs:56312
```

`ForeachTail.key_type`, `assoc_index_widths` and `assoc_index_props` are core AST / elaboration fields that exist only locally — no open PR in `aionhw/xezim-core` adds them. Until that PR is opened, nothing here can be built or evaluated by anyone else, and CI cannot help: the workflow clones `xezim-core` from **upstream main**, never a paired branch.

### What I picked (merged as `3aa55fd`)

Applied all 20 new testcases to current `main` and ran them. Four pass unmodified and are now merged, taken verbatim with credit to you:

- `tests/classes/class_unpacked_struct_property.rs`
- `tests/classes/struct_output_formal.rs`
- `tests/collections/string_foreach_content.rs`
- `tests/types/byte_local_narrow.rs`

### What did not pick, and why it is not 22 bugs

The other 15 files produce 22 failures, but these are **not** defects on `main` — they are the test half of the missing core change. The failing clusters map straight onto the three absent fields:

| Failing cluster | Missing core field |
|---|---|
| dynarray properties ×3, `foreach` dynarray | `ForeachTail.key_type` |
| assoc index narrowing ×3, class coll bits ×2, multidim assoc exists | `assoc_index_widths` / `assoc_index_props` |
| whole queue copy ×3, queue indexed grow, struct coll element | (same collection work) |
| `$bits` member shadow, randomize obj array, string formal leak ×3, bits builtin width, size on class member | mixed |

`misc/string_formal_name_leak.rs` is the one partial: `string_return_name_does_not_leak` passes, the other three fail. I left the whole file with you rather than split someone else's testcase.

### Already superseded

Four of the added files are **byte-identical** to files now on `main`, landed via #87 and #82 today:

```
tests/classes/pure_sv_phase_objection.rs
tests/classes/type_param_struct_formal.rs
tests/scheduling/procedural_loop_stall.rs
tests/scheduling/procedural_loop_stall.sv
```

The UVM shim removal is also already in (`uvm_obj_count` is gone from `main`), which is the bulk of the `-2873` in `simulator.rs` here and the source of the conflict.

### Two things to fix on a rebase

- This deletes `tests/classes/uvm_config_db_tests.rs`. That file was dead until today — it `.expect()`ed a hardcoded `../1800.2-2020.3.1` path no workflow provided — and now resolves the library properly and runs three live tests. Deleting it would remove coverage that only just became real.
- `fscanf_pos_tmp.txt`, `work/temp.txt` and `work/example.dump` are committed here. They are test outputs; `main` now ignores them (`7bd5881`).

### Suggested shape

Open the `xezim-core` PR with `ForeachTail.key_type` and the `assoc_index_*` fields first, then rebase this onto current `main` — which is a lot smaller once #87/#82 are subtracted. The remaining 16 tests become reviewable the moment their core half exists.

Thomasb81:
Requested PR is here : https://github.com/aionhw/xezim-core/pull/21
Will proceed with rebase now

Thomasb81:
Superseded by the paired release_pr PR #111 (UVM seed_map/randomization, is_active config, $cast and data-path fixes).

PR #104, opened 2026-08-08
fix: LRM-compliant rejections for parameterized class scope + untyped port net assignment

opensource-elearning:
This PR implements two additive LRM-conformance checks in the `should_fail` second-pass lint:

## 1. Reject unadorned parameterized-class scope resolution (§8.25)
- **sv-tests:** `chapter-8/8.25.1--parametrized_class_invalid_scope_resolution.sv`
- **LRM 1800-2017 §8.25:** "For a parameterized class C, the default specialization is C#(). Other than as the prefix of the scope resolution operator, use of the unadorned name of a parameterized class shall denote the default specialization."
- An unadorned parameterized class used as a `::` prefix (`par_cls::b`) does NOT denote the default specialization and is illegal; a legal form must specialize explicitly (`par_cls#()::b`, `par_cls#(15)::b`).
- The AST distinguishes the two: the legal form's `MemberAccess` base is a `Specialization`, the illegal form's base is a bare single-segment `Ident`. The lint walks every statement/continuous-assign expression and fires only when that base names a class whose HEADER `#(...)` list is non-empty (`param_order` / `type_param_names`), matching the simulator's `class_is_parameterized`. Body-only localparams do NOT parameterize a class, so UVM-style static-call initializers are untouched.

## 2. Reject procedural assignment to untyped output/inout ports (§23.2.2.3, §6.5)
- **sv-tests:** `chapter-14/14.3--clocking-block-signals-error.sv`
- **LRM §23.2.2.3:** An ANSI output/inout port with an omitted or implicit data type defaults to a NET of default net type; procedural assignment (`=`/`<=`) to such a net is illegal (§6.5). A port explicitly typed with `logic`/`var` or completed by a body variable declaration is a variable and legal.
- Extended `check_proc_net_assign` to collect untyped output/inout ANSI ports as implicit nets, excluding ports with `var_kw`, explicit `data_type`, or a body `DataDeclaration` completion (mirrors the existing port-completion logic in elaborate.rs).

## Validation
- sv-tests chapter suite: **823/830 pass** (7 remaining are SVA/constraint runtime and interface-class typedef-inheritance)
- full native suite green (**1750 tests**)
- Both checks are additive in `should_fail_lint` only — they never change existing compile/elaborate behavior, only add diagnostics for already-built AST + elaborated module.

## Native test correction
- `str_param_spec.sv` used the illegal bare `Wrapper::type_name()` form; corrected to the LRM-conformant explicit empty specialization `Wrapper#()::type_name()` which preserves the assertion intent and matches slang (which reports `GenericClassScopeResolution` for the bare form).

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -12,7 +12,7 @@ jobs:
 
       - name: Checkout xezim-core
         run: |
-          git clone https://github.com/aionhw/xezim-core.git ../xezim-core
+          git clone https://github.com/opensource-elearning/xezim-core.git ../xezim-core
           echo "CORE_SHA=$(git -C ../xezim-core rev-parse HEAD)" >> "$GITHUB_ENV"
 
       - name: Install Rust toolchain
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
 
       - name: Checkout xezim-core
         run: |
-          git clone https://github.com/aionhw/xezim-core.git ../xezim-core
+          git clone https://github.com/opensource-elearning/xezim-core.git ../xezim-core
           echo "CORE_SHA=$(git -C ../xezim-core rev-parse HEAD)" >> "$GITHUB_ENV"
 
       - name: Install Rust toolchain
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -77126,6 +77126,7 @@ impl Simulator {
             constraints.push(ClassConstraint {
                 is_static: false,
                 is_extern: false,
+                is_pure: false,
                 has_body: true,
                 name: crate::ast::Identifier {
                     name: "__inline__".to_string(),
--- a/src/should_fail_lint.rs
+++ b/src/should_fail_lint.rs
@@ -31,6 +31,9 @@ pub fn lint_should_fail(defs: &[&SourceDefinition], elab: &ElaboratedModule) -> 
     // §23.3.2: map every module/interface/program to its declared port names,
     // so a named port connection to a non-existent port can be rejected.
     let port_map = build_port_map(defs);
+    // §18.5.2: scans ALL classes at once (needs the full hierarchy to walk
+    // extends chains), so it runs once rather than per-definition.
+    check_pure_constraints(elab, &mut errs);
     for def in defs {
         match def {
             SourceDefinition::Class(c) => check_class(c, &mut errs),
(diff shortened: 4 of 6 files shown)


aionhw:
Reviewed against `aionhw/main`. **Request changes** — two blockers are independent of the code itself.

### Blocking

1. **CI is repointed at a fork.** `.github/workflows/test.yml:22` and `msrv.yml:9` change `git clone https://github.com/aionhw/xezim-core.git` to `https://github.com/opensource-elearning/xezim-core.git`. Merging this makes upstream CI build and test every future PR against a third-party repository. This isn't mentioned in the PR description.

2. **The PR doesn't compile against `aionhw/xezim-core`** — which is why (1) is there. `check_pure_constraints` reads `ClassConstraint::is_pure` and `check_specparam_use` calls `def.specparams()`; neither exists in core today (`xezim-parser/src/ast/decl.rs:890` has `{ is_static, is_extern, has_body, name, items, span }`, and there is no `fn specparams` anywhere). Please land the core-side change first and repoint CI back to `aionhw`.

3. **"Additive, never changes existing compile/elaborate behavior" isn't accurate.** `lint_should_fail` is called unconditionally on the simulate path (`src/lib.rs:748`, `src/main.rs:2382`) and a non-empty result becomes a hard error. Every new check is a new way to reject a design at *simulate* time, not only under `--compile`.

4. **Scope exceeds the description.** The body describes two checks; the diff adds five behavior changes (parameterized-class scope, untyped-port net assign, specparam-in-parameter, block-local enum tracking + compound assign, pure-constraint enforcement). The specparam and pure-constraint checks are undescribed. Please split.

### Rejects-valid risks (all untested)

5. **The untyped-port check has no tests at all** — neither the rejection nor a still-accepted case (`src/should_fail_lint.rs`, the `PortList::Ansi(ps)` block). It runs against every ANSI port list in the design; this is the highest-risk check here.

6. **Body-completion exclusion is incomplete.** The `vars` set collects only `ModuleItem::DataDeclaration`, but a body redeclaration written `output logic q;` is a `ModuleItem::PortDeclaration`. That form now rejects. `tests/scheduling/waiter_edge_ordering.rs:25` only survives because it happens to spell it `reg [7:0] q;`.

7. **`check_scope_expr` fires on `.` as well as `::`.** The parser builds `ExprKind::MemberAccess` for both. The check consults the flat global `elab.classes` (which also registers nested classes under their bare inner name), guarded only by `elab.signals.contains_key(base)` — and `elab.signals` is keyed by flattened hierarchical name, so a same-named signal in a non-top module doesn't shadow. Any `X.field` where `X` collides with a parameterized class name anywhere in the design is wrongly rejected.

8. **§8.25 exception unhandled**: inside a generic class's own body the unadorned class name as a `::` prefix is legal. Latent today (the walk doesn't reach class bodies), but the doc comment states the rule unconditionally.

9. **Block-local enum vars go into one module-wide set**, then assignments are checked across all module items — so an enum `val` in one `initial` makes `val = 1` in a different block (or at module scope, where `val` may be a distinct `int`) an error. Common names make this plausible.

10. `check_pure_constraints` collects `implemented` order-insensitively: base implements `c` → virtual middle redeclares `pure constraint c` → concrete derived omits it, and it passes. Also emits duplicate errors for nested classes (two keys in `elab.classes`).

### Nits
- `tests/types/ivtest_cast_cluster.rs`: dead `let _ = SRC;` and a "no assertion body needed" comment directly above an assertion.
- The §23.2.2.1 citation for the body-completion exclusion is off — an ANSI port may not legally be redeclared in the body at all; the leniency is what keeps the suite green, so say that.
- `tests/lrm_9_value_param/str_param_spec.sv:30` (`Wrapper::type_name()` → `Wrapper#()::type_name()`) is correct per §8.25, but it shows in-repo code already used the bare form — worth checking the UVM corpus before enabling the check.

opensource-elearning:

aionhw:
Closing in favor of #134 — same two checks resubmitted there by the author, and this branch conflicts with main on `msrv.yml`. Review continues on #134.

Issue #24, opened 2026-07-12
Data to String Formatting Functions ($swrite(b) / $sformat(f)) Behavior


Ganesh-TNL:
Sample test case on EDA Playground for LRM section 21.3.1 and 21.3.3 (formatting data to string)

https://www.edaplayground.com/x/NXYe

xezim 0.8.5 fails basic sformat and swrite checks. 

```text
[ERROR] Basic $sformat assignment failed! Expected: "Value is 42 and status is OK", Got: "xezim"
[ERROR] Literal $sformat copy failed! Expected: "Static text with no placeholders", Got: "xezim"
[ERROR] Extraneous checker isolated via specifier filtering failed! Expected: "Only want 99", Got: "xezim"
[ERROR] Basic $swrite non-spaced serialization failed! Expected: "Part A: 10 | Part B: 15", Got: "xezim"
[ERROR] Interleaved multi-format string streams failed! Expected: "Hex a5 Bin 1100 Done.", Got: "xezim"
[ERROR] Self-referencing string extension buffer failed! Expected: "Initial Appended 42", Got: "Initial "
[ERROR] $swriteb implicit fallback vector parsing failed! Expected: "1010", Got: "Initial "
[ERROR] $swriteb mixed specifier and direct dump parsing failed! Expected: "Val: 111 Hex: ff", Got: "Initial "
```

<details>

<summary>Output log from xezim 0.8.5 for the compliance test case</summary>

```text
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim data.to.string.fmt.sv > log.data.to.string.fmt.log 2> /dev/null
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ cat log.data.to.string.fmt.log
=== xezim 0.8.5 ===
Max time: 100000
------------------------------
==========================================================
Starting Unified LRM 21.3 String Formatting Test Suite
==========================================================
[PASS] Baseline Test (Direct String) passed.
[PASS] Baseline Test (Direct Variable) passed.

--- Testing Section 21.3.3: String Buffering Tasks ---
[ERROR] Basic $sformat assignment failed! Expected: "Value is 42 and status is OK", Got: "xezim"
[ERROR] Literal $sformat copy failed! Expected: "Static text with no placeholders", Got: "xezim"
Hex: 5a, Bin: 1100
[PASS] Basic $sformatf assignment passed.
[PASS] Inline $sformatf evaluation passed.
[...]

PR #112, opened 2026-08-11
xezim: class/$cast/collection/event semantics + factory re-entrancy guard + regression tests

Thomasb81:
## Summary

`xezim` simulator runtime correctness fixes + pure-SV regression tests on the `release_pr` branch. Paired with the **`xezim-core`** companion PR **#24** ([aionhw/xezim-core#24](https://github.com/aionhw/xezim-core/pull/24)), which provides the parser/elaboration changes this PR depends on (the `EventTrigger.target` AST field and the elaboration of class-typed associative-array element types). Both PRs should land together.

This PR builds cleanly only against the `xezim-core` `release_pr` branch; the `next` branch here points `xezim-core` at `../xezim-core` (path dependency).

## Changes vs `main` (20 commits)

The bulk of the runtime work is in `src/compiler/simulator.rs` (~1900 lines). The large net deletion is the removal of a stray `simulator.rs.orig` backup artifact that had been committed by accident.

### Class / object semantics
- **Static method dispatch through object handles** — `obj.get_type()` / `obj.Type::method()` now resolve through the handle's class instead of dropping the call.
- **`$cast` correctness (§8.25)** — `$cast` to one's own `this_type` inside a parameterized class succeeds; `$cast` between two *distinct* parameterized-class specializations (differing type params) now correctly fails.
- **Active specialization wins** when resolving a colliding type parameter, so a stale cached binding no longer shadows the active specialization's own argument.
- **Class-method formal widening** — a class-method formal is widened to a typedef'd integral's full width on bind.
- **`new(args)` runs the constructor** for a parenthesized `new` with a class-formal (was treated as a shallow copy); complements the parser-side `new` disambiguation in xezim-core #24.
- **Bare `obj.randomize`** (no parens) is dispatched as a parameterized call.
- **Whole-struct copy / local-decl init from a class property** keeps handle and string members (no longer dropped or shallow-copied as raw bits).

### Associative arrays & collections
- **Class-typed package-scope assoc `= new`** stores a real instance; complements the xezim-core #24 element-`DataType` recording.
- **`f().member` assoc writes/reads** — writes/reads to an associative-array member reached through a *function-returned* handle (`f().member[k] = v`) now resolve the member instead of being silently dropped.
- **Factory `register()` re-entrancy guard** — a `register(obj)` whose first statement `obj.get_type_name()` triggers a parameterized class specialization's lazy static init (which re-enters `register(obj)` for the same object) is guarded so the object is registered exactly once, instead of emitting a spurious "already registered" warning when the outer call resumes.
- **Assoc keys containing dots** are counted correctly by `.num()` (they were split as hierarchical names).
- **Assoc-index narrowing** (§7.8.2) — a wider index expression is narrowed to the declared index type before keying; complements xezim-core #24.

### Events & concurrency
- **`@ev` through an array element suspends** — `@(arr[k].ev)` / `@(m_events[obj].all_dropped)` no longer wake immediately at t=0; the receiver is evaluated at suspension time and matched against a declared `event` property.
- **`-> arr[k].ev` round-trips** — the trigger now fires the same instance-event identity the wait-side suspends on (complements the parser Index-flattening fix in xezim-core #24).
- **`disable fork` drops a killed pid's pending timer**, so a killed child no longer fires a deferred wakeup.

### Misc
- **`$bits` / `$size`** builtin width on class members and integral types (complements xezim-core #24 const-eval).
- Removed the committed `simulator.rs.orig` backup artifact.

### Regression tests (pure SV, ~3000 lines added)
New `tests/{classes,collections,misc,types}` modules covering: dynamic-array/dynarray property element new/copy/size, assoc-index narrowing, class-collection bits/pattern, queue/struct element copy, assoc class-new stores instance, dotted assoc keys, bare method-call return, bare randomize, function-returned-handle assoc member, factory re-entrant register, indexed-event round-trip, disable-fork timer drop, new-ctor vs shallow-copy, param-class `$cast` (own + differing), static-method-via-handle, string-formal scope leak, `$bits`/`$size` on class members.

## Test plan
- `cargo build` clean (against xezim-core `release_pr`).
- Full self-test suite: **1935 passed, 0 failed**.
- A Heisenberg control (guard disabled) restores the spurious registrations, confirming the factory re-entrancy fix is semantic rather than a layout artifact.

## Depends on
- **xezim-core #24** ([aionhw/xezim-core#24](https://github.com/aionhw/xezim-core/pull/24)) — parser `new`/event-trigger + elaboration fixes.


aionhw:
Thanks Tom — this PR's content has been landed on main via cherry-pick (through 2b2ad21), together with the companion core changes (xezim-core through 6e74592). Summary of what happened to each piece:

**Landed as-is (cherry-picked with `-x` back-references, your authorship preserved):** static method dispatch through handles, both `$cast` specialization fixes, active-spec-wins, class-method formal widening, `new(args)` ctor-vs-shallow-copy (with the core `ShallowCopy` parser change), bare `obj.randomize` dispatch, class-typed package-assoc `= new`, `f().member` assoc access, the factory register re-entrancy guard, the AA index-narrowing family (merged with main's signed-key handling so negative keys keep working at the declared width), the disable-fork/proc_kill timer purge (applied to all four kill loops), the typedef-extends heritage walk, `simulator.rs.orig` removal, and ~25 of your tests.

**Dropped as duplicates of main:** the dotted-assoc-key filter removal (main 5392d53), the whole-struct-copy fix (main 0dedf70 — note its ternary picks the else-arm on an X condition where your version fell back to the packed path), the alias-slot port, and the Cargo.toml path-dep (main now uses a git dependency; see README "Modifying xezim-core").

**Deferred (worth their own PR):** the ipc/oop/process module split — review verified it move-equivalent (nice work), but re-applying main's recent scheduler/assoc work into the new layout deserves its own focused change; and b823d8d's frame-scoped `var_class_types`/`string_signals` (+ its test), which is entangled with that split.

**Repairs made while landing:** `uvm_config_db_tests.rs` was restored (its deletion rationale referenced a symbol the file never used — and per the new AGENTS.md rule, test retirements need mapped successors); `unpacked_struct_class_property_whole_value` was re-registered; one doc comment's vendor-simulator name was anonymized; `randomize_obj_array_property` got constraint pinning (raw randomize + `!= 0` asserts had a 2^-32 flake); `class_struct_copy.rs` (root-level binary, redundant with main's `struct_prop_whole_copy`) was not taken.

**Two follow-ups flagged from review, no action needed now:** the register guard evaluates a side-effecting first arg twice for any class with a `register` method (consider gating on the factory class), and `expr_assoc_name` re-evaluates `f()` per probe for `f().member` receivers.

Core side (#24): commits landed with two adjustments — the new serialized fields/variants are appended last (bincode index stability) and `XEZIM_BYTECODE_MAGIC` bumped 0x12→0x13; `assoc_key_types` was skipped as main already had that field. Closing both PRs as landed.

PR #115, opened 2026-08-13
simulator: foreach over class localparam fixed-size arrays

chenbhil:
Class `localparam T Arr [N]` declarations are `is_static = true`, so they're excluded from `array_properties`/`module.arrays` (the `!is_static` guard in elaborate.rs) and no shape was recorded for the async ForeachTail path. foreach_materialize_keys_1d returned None, which exec_park_cont reads as "not ready" and replays the same iteration forever (i=0 repeating) instead of completing.

Three call sites needed the new shape source:
- the synchronous foreach-array-iterate branch now goes through foreach_dims() instead of module.arrays directly, so it picks up class localparam arrays too
- the async ForeachTail path gets a new branch that walks the runtime class (via the heap, not class_context_stack) up the inheritance chain looking for a localparam_array_dims entry
- foreach_dims() itself grows a fallback to the same lookup, for callers (e.g. $size on a class-scoped name) that go through it directly

Depends on the companion xezim-core change that adds ElaboratedClass::localparam_array_dims (branch
fix/foreach-localparam-array on xezim-core) - CI on this PR is expected to be red until that one merges.

Repro: repros/localparam_foreach_repro.sv (foreach must reach i=0,1,2 instead of stalling on i=0 forever).

Note: this fixes foreach ITERATION only. A separate, still-open bug found while re-verifying this fix: reading an ELEMENT VALUE out of a class-scoped localparam array (e.g. `Arr[i]` in an ordinary expression, not just as a foreach target) returns 0 regardless of index - see repros/class_localparam_array_read_repro.sv. Not fixed by this change. will file separately.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -49238,8 +49238,15 @@ impl Simulator {
                                 }
                                 self.continue_flag = false;
                             }
-                        } else if let Some(&(lo, hi, _)) = self.module.arrays.get(&name) {
-                            // Fixed unpacked array: iterate its declared index range.
+                        } else if let Some(dims_1d) = self
+                            .foreach_dims(&name)
+                            .filter(|d| d.len() == 1)
+                        {
+                            // Fixed 1-D unpacked array (module-scope or class
+                            // localparam): iterate its declared index range.
+                            // foreach_dims covers both module.arrays and
+                            // ElaboratedClass::localparam_array_dims.
+                            let (lo, hi) = dims_1d[0];
                             let descending = self.module.descending_arrays.contains(&name);
                             let mut idx = lo;
                             while idx <= hi {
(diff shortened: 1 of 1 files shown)


aionhw:
Superseded — the root cause turned out to sit upstream of `foreach`, so the fix landed in `xezim-core` (`12467b3`) instead of here.

### Why the stated premise doesn't hold

The description has class localparams as `is_static = true` and excluded from `array_properties`/`module.arrays` by the `!is_static` guard in `elaborate.rs`. They never reach that guard: a class-body `localparam` parses as `ClassItem::Parameter`, not a `ClassProperty`, and takes a separate elaboration path entirely.

What actually happened on that path: each parameter was stored as a single scalar `Signal` sized by `resolve_type_width`, and the declarator's unpacked dimension was dropped. `localparam int Arr[3]` became one 32-bit scalar.

That single omission produced four symptoms, all silent:

| | before | after |
|---|---|---|
| `Arr[i]` | `0` — a *bit* select of the scalar | `10 20 30` |
| `$size(Arr)` | `32` (the bit width; `96` for a 3-deep one) | `3` |
| sync `foreach` | 1 iteration | 3 |
| blocking-body `foreach` | **0** iterations | 3 |

Rows 1–2 are the element-value bug noted here as separate and still-open (`class_localparam_array_read_repro.sv`). It is the same bug. That's the main reason this PR wasn't merged as-is: fixing shape alone makes `foreach` iterate correctly over zeros, so the visible result stays wrong.

### The fix

Register the dimension and the initializer the way a `const` array property is registered — a path that already worked end-to-end for the identical shape. Consequences:

- no changes to `simulator.rs` at all (this PR: +73/−6 across three call sites)
- no new field on `ElaboratedClass`, so no `XEZIM_BYTECODE_MAGIC` bump

On that last point, worth flagging for the companion core branch: it adds a serialized field to `ElaboratedClass`, and per core's `AGENTS.md` that requires bumping the magic's version byte plus a version-ladder line — otherwise stale `.xez` artifacts deserialize garbage instead of failing with "recompile with current xezim".

Regression tests are in xezim `0447da3` (`tests/classes/class_localparam_array.rs`): element reads, `$size`, both `foreach` paths, and a dimension sized by an earlier class localparam.

One parsing gotcha found on the way, for anyone in this area later: `[N]` with a bare identifier is grammatically ambiguous — a type name means associative, a constant means a size — and the parser always emits `UnpackedDimension::Associative`. The localparam path now treats an identifier that const-evaluates as a size. The same ambiguity means a plain property `int Prop[N]` still elaborates as associative, which is a separate pre-existing limitation.

### The part of this PR worth keeping

The failure-mode diagnosis was right and is the more valuable finding. `foreach_materialize_keys_1d` returning `None` makes `exec_park_cont` replay the same iteration forever — a silent hang with no diagnostic.

`XEZIM_STALL_LIMIT` does **not** catch it: that guard is per-timestamp, and a `foreach` body blocking on a clock edge advances time on every replay, so it sails straight past. A no-progress guard on the `ForeachTail` continuation — same `idx` replaying past a threshold, hard error naming the array — would turn this whole class of bug into a one-line diagnostic instead of a multi-hour hang. Not implemented yet; worth a follow-up.

Also still open and separate: mutable `static int S[3]` in a class reads garbage (`0 0 1`, `$size` 96). Different declaration path, and it needs real shared storage — per-instance copies would be semantically wrong for a mutable static.

chenbhil:
Confirmed.
Closing as superseded by xezim-core@12467b3. Thanks for the root-cause reframe.
Glad the failure-mode diagnosis was useful. I see the no-progress replay guard landed in 264f7ab. That silent hang was the worst part to track down.

Issue #26, opened 2026-07-13
Static Variable Initialization with System Functions


rharikrishna25:
The following SystemVerilog test case fails with xezim v0.9.0:

<details>

<summary>SV Code with Static Variables Initialized using System Functions ( sv_static_init_tb.sv )</summary>

```systemverilog

`ifndef SVTEST_DEFS_SVH
`define SVTEST_DEFS_SVH

`define SVTEST_INIT \
  int failures = 0;

`define SVTEST_CHECK(expr, msg) \
  if (!(expr)) begin \
    failures++; \
    $display("FAIL: %s", msg); \
  end

`define SVTEST_PASSFAIL \
  if (failures == 0) begin \
    $display("TEST_PASS"); \
  end else begin \
    $display("TEST_FAIL count=%0d", failures); \
    $fatal(1); \
  end
`endif

module sv_static_init_tb;

  // 1. Math / Dimension Functions (Elaboration-safe)
  static int         math_clog2 = $clog2(32);
  static real        math_sqrt  = $sqrt(16.0);
  static int         array_size = $size(logic [7:0]);

  // 2. String & Hierarchy Functions (Simulation-safe)
  // Note: $sformatf("%m") captures the exact instance path at time zero


  static string      inst_path  = $sformatf("Path is %m");
  static string      type_name  = $typename(int);

  // 3. Command Line / Environment Functions (Simulation-safe)
  // To test plusargs, run simulation with: +TEST_MODE=1 +SEED_VAL=42
  static int         has_plusarg = $test$plusargs("TEST_MODE");
  static int         plusarg_val = get_plusarg_seed();

  // Helper function to extract a plusarg into a static variable at initialization
  function static int get_plusarg_seed();
    int val;
    if ($value$plusargs("SEED_VAL=%d", val)) begin
      return val;
    end
    return 0; // Default if not found
  endfunction

  // 4. Randomization Functions (Simulation-safe)
  static int         rand_init  = $urandom_range(100, 200);

  // -------------------------------------------------------------
  // Test Execution Block
  // -------------------------------------------------------------
  initial begin
    // Instantiate your macro to create the 'failures' counter
    `SVTEST_INIT

    $display("--- Starting Static Initialization Verification ---");

    // Check Math and Dimension initializations
    `SVTEST_CHECK(math_clog2 == 5,    "math_clog2 should be 5")
    `SVTEST_CHECK(math_sqrt == 4.0,   "math_sqrt should be 4.0")
[...]

PR #116, opened 2026-08-14
sim: struct-formal binding, foreach-over-string-collections, and randomize-solver fixes

Thomasb81:
## Summary

Simulator (pure-SystemVerilog) fixes and self-tests for class **struct-formal**
binding, **foreach over string (assoc) collections**, the constraint solver's
handling of `randomize()`/`randomize() with {}` inline blocks, dyn-array
`.size()` to co-constrained-scalar coupling, `static` locals across
re-entrancy, and suspended `process::await()`-parked fork children.

Branch was rebased onto the fork's current `main`, so this PR presents only
the feature delta (26 files, +2516/-111) against `main`.

## Simulator changes (`src/compiler/simulator.rs`)

- **Bare `randomize()` in a class method** - route unqualified
  `randomize()`/`srandom()`/`get_randstate()`/`set_randstate()` inside a class
  method through the 18.11 solver (implicit built-in dispatch) so constructors
  honour declared constraints instead of leaving `rand` fields at defaults.
- **`obj.randomize() with {}` inline blocks** -
  force prefixed inline members (`obj.member == K`) in the Eq solver, keyed by
  the in-flight receiver name; resolve enclosing-scope (caller) members by
  freezing them to their current values before the solver frame; thread the
  randomized object's handle through the coupled-affine solve.
- **dyn-array `.size()` to co-constrained scalar** - size a rand dynamic array
  after scalar-equality propagation and treat a dyn-array `.size()` operand as
  unresolved when coupling `arr.size() == scalar` to the array's length.
- **static locals** - share a static subroutine local across re-entrant/
  recursive activations (single live cell, LRM 6.21/13.3.1) and key its
  persistent cell by the declaring subroutine.
- **process suspension** - keep a `process::await()`-parked fork child
  suspended (don't reap it); inline parameterized static blocking calls
  (`C#(T)::method`) into the suspend-aware stream with `Specialization`
  receivers awaiting on `@event`.
- **get_type()** - resolve a started sequence's member static type.
- **assoc collections** - iteration keeps full string keys containing `[ ]`;
  multidim assoc recursion-guard and type-param-typed assoc-pool fixes.

## New tests

Self-contained pure-SystemVerilog regressions, validated byte-for-byte against
the reference simulator: struct_formal_and_config_foreach,
constraint_dyn_size_pinned_scalar, constraint_inline_enclosing_scope,
constraint_prefixed_inline_with, bare_randomize_in_method,
static_recursion_shared_cell, phase_jump_static_latch,
fork_join_none_await_context, determinism_and_stall, assoc_bracket_keys,
multidim_assoc_struct_copy, typeparam_pool_wait, param_static_park, plus
uvm-integration additions.

Scheduling / misc / types / classes suites keep passing.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -48097,6 +48097,103 @@ impl Simulator {
                         while let ExprKind::Index { expr: inner, .. } = &root.kind {
                             root = inner.as_ref();
                         }
+                        // §4c/: `obj.member[k] = new(...)` — the collection is
+                        // a MEMBER of a class handle (`mid.base[i] = new(...)`
+                        // in do_unpack, or the flattened nested
+                        // `a.mid.base[i] = new(...)`). The root unwraps to a
+                        // MemberAccess or a flattened >=3-seg Ident. Resolve
+                        // the element class from the RECEIVER's runtime class
+                        // property so the bare `new(...)` constructs the
+                        // declared element type instead of the enclosing
+                        // `this` class (which would run a component ctor and
+                        // trip the UVM ILLCRT guard deep inside the library).
+                        let recv_member: Option<(Expression, String)> = match &root.kind {
+                            ExprKind::MemberAccess { expr: recv, member } => {
+                                Some(((**recv).clone(), member.name.clone()))
+                            }
+                            ExprKind::Ident(ih) if ih.path.len() >= 3 => {
+                                let mut p = ih.clone();
+                                let member = p.path.pop().unwrap().name.name.clone();
+                                p.cached_signal_id = std::cell::Cell::new(None);
+                                Some((Expression::new(ExprKind::Ident(p), lvalue.span), member))
+                            }
+                            _ => None,
+                        };
+                        if let Some((recv, member_name)) = recv_member {
+                            let this_recv = matches!(&recv.kind, ExprKind::Ident(rh)
+                                if rh.path.len() == 1 && rh.path[0].name.name == "this");
+                            let handle = if this_recv {
+                                self.this_stack.last().copied().flatten().unwrap_or(0)
+                            } else {
+                                self.eval_handle_expr(&recv).unwrap_or(0)
+                            };
+                            if handle != 0 {
+                                let hcn = self
+                                    .heap
+                                    .get(handle)
+                                    .and_then(|o| o.as_ref())
+                                    .map(|i| i.class_name.clone());
+                                let mut mtn: Option<String> = None;
+                                let mut cur = hcn;
+                                while let Some(cn) = cur {
+                                    if let Some(cd) = self.module.classes.get(&cn) {
+                                        if let Some(tn) = cd
+                                            .properties
+                                            .get(&member_name)
+                                            .and_then(|s| s.type_name.clone())
+                                        {
+                                            mtn = Some(tn);
+                                            break;
+                                        }
+                                        cur = cd.extends.clone();
+                                    } else {
+                                        break;
+                                    }
+                                }
+                                if let Some(mcls) = mtn
+                                    .as_deref()
+                                    .and_then(|t| {
+                                        self.resolve_typedef_spec(t)
+                                            .map(|(b, s)| (b.clone(), Some((b, s))))
+                                            .or_else(|| {
+                                                self.resolve_simple_typedef_class(t)
+                                                    .filter(|c| self.module.classes.contains_key(c))
+                                                    .map(|c| (c, None))
+                                            })
+                                            .or_else(|| {
+                                                self.extract_spec_from_string(t)
+                                                    .filter(|(b, _)| self.module.classes.contains_key(b))
+                                                    .map(|(b, s)| (b.clone(), Some((b, s))))
+                                            })
+                                            .or_else(|| {
+                                                self.module
+                                                    .classes
+                                                    .contains_key(t)
+                                                    .then(|| (t.to_string(), None))
+                                            })
+                                    })
+                                {
+                                    let ctor_args: &[Expression] = match &rvalue.kind {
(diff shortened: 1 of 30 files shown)


aionhw:
Thanks for this — the static-locals work is fixing something real, and I want to flag that up front: for a recursive `static`, `main` today returns the wrong answer and this PR returns the right one.

```systemverilog
function automatic int rec(int n);
  static int depth = 0;
  depth = depth + 1;
  if (n > 0) return rec(n - 1);
  return depth;
endfunction
// rec(3): reference simulator = 4, xezim main = 1 (wrong), this PR = 4 ✔
```

## One blocking issue

The same change binds an inner subroutine's **plain automatic** local to an ancestor's `static` of the same name, in both directions:

```systemverilog
function automatic int inner();
  int cnt = 99;            // ordinary automatic local
  cnt = cnt + 1;
  return cnt;              // must be 100
endfunction
function automatic int outer();
  static int cnt = 5;
  return inner() * 1000 + cnt;   // must be 100005, every call
endfunction
```

| | 1st call | 2nd call |
|---|---|---|
| reference simulator | 100005 | 100005 |
| xezim `main` | 100005 | 100005 |
| this PR | **6006** | **7007** |

`inner`'s local reads `outer`'s static (5 instead of 99), and `inner`'s write then corrupts that static, so the error compounds across calls. Per §13.4.2 a static local is one cell shared across calls of *its own* subroutine and nothing else.

This is a silent wrong-answer bug, and names like `cnt` / `i` / `idx` collide constantly, so nothing about the source looks suspicious. The full test suite is green on this PR, so nothing currently catches it.

Cause, as far as I can tell: `static_local_key_for` scans the frame stack in reverse and **skips frames that don't declare the name**, continuing outward, so an innermost frame holding a same-named automatic local is stepped over. The `syncs` tuple it scans already carries the correct key — returning that directly instead of re-deriving one should fix it and drop a per-read `format!` allocation on the way.

I've pushed a regression test to `main` (`tests/misc/static_local_scope_isolation.rs`) that pins this. It also pins the converse, so a fix can't work by simply giving up sharing: sibling subroutines' same-named statics stay independent, a static still accumulates across calls of its own subroutine, and a task's static coexists with a function's same-named local. It carries an `#[ignore]`d pin for the recursion case above, ready to enable with your fix.

## Two smaller things

1. **Simulator brand name in five new places** — one comment in `src/compiler/simulator.rs` and four in the new tests (`tests/scheduling/bare_randomize_in_method.rs`, `phase_jump_static_latch.rs`, `static_recursion_shared_cell.rs` + its `.sv`). House style is "the reference simulator" / "a golden simulator", which your own `tests/sv_ref/param_static_park.sv` already uses. Nice that the PR *removes* several such references elsewhere — this is just the other half.

2. **`tests/scheduling/phase_jump_static_latch.sv`** is a 216-line verbatim third-party test carrying three vendors' copyright headers. Apache-2.0 with the notice retained, so redistribution is fine, but no file in this repo currently carries vendor copyright — that's a precedent for the repo owner to set deliberately rather than absorb in a large PR. A hand-written equivalent exercising the same phase-jump/static-latch behaviour would avoid the question entirely.

## Everything else checked out

Merges cleanly onto current `main`; release, JIT and debug suites all green with `aionhw/xezim-core#26` applied; the five modified existing tests change comments only, with no assertion weakened; the `timeout 60` in `tests/sv_ref/run.sh` is a good call. `xezim-core#26` looks sound to me and I'd take it as-is.

Two things I did not verify but that might be worth a second look, purely as questions: the randomize pipeline now runs collection sizing *after* the scalar fixpoint, which affects every `randomize()` with a rand collection rather than only the new coupled-size path — is a scalar pinned from an element (`first == payload[0]`) still solved before the elements are redrawn? And is `await_waiters` purged when the *waiting* process is killed, or only when its target is?

aionhw:
Landed on `main` — thanks for this. Your commits are merged with authorship intact (`0182a09`), so this PR shows as closed rather than merged only because the branch was integrated directly along with a follow-up fix.

**What changed on the way in**

`static_local_key_for` now resolves a static local only in the frame that declared it, and reuses the key recorded at declaration time instead of re-deriving one (`49c142a`):

```rust
self.static_local_syncs
    .last()?
    .1
    .iter()
    .find(|(n, _)| n == nm)
    .map(|(_, key)| key.clone())
```

That closes the shadowing case from the earlier review — a callee's ordinary automatic local no longer binds to an ancestor's same-named `static` — and it also removes a second, quieter failure mode: the re-derived key could drift from the declaration-time key whenever the class or specialization context differed between declaration and use, so a static could be looked up under a name it was never stored under. As a side effect it drops a per-read `format!` from a path that runs for every bare identifier once any static exists.

Your recursion fix is untouched and still correct — `rec(3)` returns 4, where `main` previously returned 1. `static_recursion_shared_cell`, `phase_jump_static_latch` and `bare_randomize_in_method` all pass unchanged, and `tests/misc/static_local_scope_isolation.rs` (pushed earlier as `f790ecf`) pins both directions so neither can regress: the shadowing case, and the sharing your change restored.

I also replaced the simulator brand name with "the reference simulator" in the five new comments, matching house style and what your own `tests/sv_ref/param_static_park.sv` already does.

**Verification**

Release, JIT and debug suites all green: misc 754, classes 263, scheduling 301, strings 111, types 334, hierarchy 127. I separately confirmed this does **not** depend on `aionhw/xezim-core#26` — the full suite passes against the currently pinned core — so no rev-pin bump was needed and that PR stands on its own. It's still open and reviewed favourably.

The two questions from my earlier comment (collection sizing now running after the scalar fixpoint, and whether `await_waiters` is purged when the *waiting* process is killed) are still worth a look, but neither is blocking and nothing in the suite exercises them today.

PR #120, opened 2026-08-22
Enable the §6.6.7 resolution tests; add an EEnet RNM regression

vrajeshprakhya:
# Enable the §6.6.7 resolution tests; add an EEnet RNM regression

**Draft until the companion core PR merges.** Depends on: https://github.com/aionhw/xezim-core/pull/27
Issue: https://github.com/aionhw/xezim/issues/119

The elaborator changes live in `xezim-core`. Once that PR merges I'll bump both
`rev = ...` fields in `Cargo.toml` to the merged SHA and mark this ready — until
then this branch does not build against the pinned core.

Two commits, +168 / −11.

## Commit 1 — `Restore is_real on a real member projected from a call result`

`eval_unpacked_call_member` sliced the member out of the packed return value but
dropped the per-leaf `is_real` flag that `struct_leaf_layout` had **already
computed** (it was bound to `_`). A `real` member read out of any struct-returning
function came back as its raw f64 bit pattern:

```systemverilog
typedef struct { real f1; bit f2; } T;
function automatic T mk(); ... endfunction
$display("%f", mk().f1);   // 4609434218613702656.0, want 1.5
```

Not nettype-specific — it affects any struct-returning function — but it is the
read path a resolution function's struct return goes through, so it also blocked
struct-of-real nettypes. Separated into its own commit; it stands alone and is
verified green against released core (2241 passed, 0 failed).

## Commit 2 — `Enable the §6.6.7 resolution tests; add an EEnet RNM regression`

The four resolution-tier tests added alongside 31 were **never runnable**: each
includes `"svtest_defs.svh"` while the harness lives at
`"../common/svtest_defs.svh"` (31 has the correct path), so all four failed in
the preprocessor before reaching simulation. Being `#[ignore]`d meant nothing
surfaced it.

- Fix the include path in 36–40; un-ignore 37–40.
- **36 stays ignored** — tagged union patterns are still genuinely unimplemented.
- Worth noting for review: **37 and 39 needed no code change.** With the include
  fixed they pass against current `main` — built-in net resolution is already
  implemented and those two were stale-ignored. Their file comments predicting
  `wand [0,1] -> 1  WRONG` are outdated; left alone here to keep the diff
  minimal, happy to update them if you'd prefer.
- Only 38 and 40 needed the core work. Both were `TEST_FAIL count=7`.

### New: `44_eenet_rnm.sv`

The "EEnet" pattern from the DVCon Europe 2019 TI/Cadence paper *Enabling Digital
Mixed-Signal Verification of Loading Effects in Power Regulation using
SystemVerilog User-Defined Nettype*. It covers the combination that motivates
UDNs in practice, which 38 (scalar resolvers) and 40 (the LRM's `Tsum`) do not:

- a struct of **three reals** (V/I/R) resolved field-by-field;
- arithmetic over the driver set no fold can express — Millman's theorem,
  `V = (Σ Vk/Rk + Σ Ik) / (Σ 1/Rk)`;
- drivers differing in **nature** — voltage sources, current sinks and passive
  loads sharing one node;
- **2, 3 and 4** simultaneous drivers;
- KCL current summing on a 1.8 V rail with three loads.

Reference values are hand-derived circuit results, so a resolver that is skipped,
folded, or called with a truncated driver set cannot pass.

## Testing

| Configuration | Result |
|---|---|
| commit 1 alone + core @ `6e2f17c` (released) | 2241 passed, 0 failed |
| both commits + companion core PR | 2246 passed, 0 failed |
| both commits + companion core PR, `--features jit` | 2250 passed, 0 failed |

2241 → 2246 is +4 un-ignored resolution tests +1 new EEnet test; ignored 15 → 11.
The real-UVM integration tests (RAL frontdoor/backdoor, TLM bench, 1.2 /
1800.2-2017 / 1800.2-2020) stay green throughout.

## Note

Nettype **ports across module hierarchy** — slide 7 of that paper, where several
instances drive a shared node through module ports — are still unimplemented.
Everything here works within a single module scope. Happy to follow up.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -51293,8 +51293,32 @@ impl Simulator {
                                     .find(|(m, _, _)| m == &member.name)
                                     .cloned()
                                 {
-                                    return base_val
+                                    let mut v = base_val
                                         .range_select((off + w - 1) as usize, off as usize);
+                                    // §6.6.7 / §7.2.1: `packed_agg_layout` is a
+                                    // 3-tuple with no is_real, so a `real`
+                                    // member projected out of a struct VALUE
+                                    // (a call result — notably a nettype
+                                    // resolution function's return) came back
+                                    // as the raw f64 BIT PATTERN: 1.5 read as
+                                    // 4609434218613702656. Restore is_real
+                                    // from the member's declared type.
+                                    let mreal = su.members.iter().any(|m| {
+                                        m.declarators
+                                            .iter()
+                                            .any(|d| d.name.name == member.name)
+                                            && matches!(
+                                                Self::resolve_type_ref(
+                                                    &m.data_type,
+                                                    &self.module.typedef_types
+                                                ),
+                                                DataType::Real { .. }
+                                            )
+                                    });
+                                    if mreal {
+                                        v.is_real = true;
+                                    }
+                                    return v;
                                 }
                             }
                         }
@@ -76165,12 +76189,21 @@ impl Simulator {
         let su = self.unpacked_struct_of(&ret)?;
         suffix.reverse();
         let want = format!("X{}", suffix.concat());
-        let (_, off, w, _) = self
+        let (_, off, w, leaf_is_real) = self
             .struct_leaf_layout("X", &su)
             .into_iter()
             .find(|(k, ..)| *k == want)?;
         let v = self.eval_expr(cur);
-        Some(v.range_select((off + w - 1) as usize, off as usize))
+        let mut out = v.range_select((off + w - 1) as usize, off as usize);
+        // `struct_leaf_layout` already knows which leaves are `real`; that flag
+        // was dropped here, so a real member projected out of a call result
+        // came back as the raw f64 BIT PATTERN (1.5 read as
+        // 4609434218613702656). This is the read path a §6.6.7 resolution
+        // function's struct return goes through — `Tsum('{...}).field1`.
+        if leaf_is_real {
+            out.is_real = true;
+        }
+        Some(out)
     }
 
     fn chain_base_packed_struct(
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -260,25 +260,25 @@ fn test_sv_35_disable_fork() {
 fn test_sv_36_tagged_union_patterns() {
     run_positive_compliance_test("tests_advanced", "36_tagged_union_patterns.sv");
 }
-#[ignore = "unimplemented feature (was dormant/unregistered before PR #48); un-ignore when implemented"]
 #[test]
 fn test_sv_37_z_skip_resolution() {
     run_positive_compliance_test("tests_advanced", "37_z_skip_resolution.sv");
 }
-#[ignore = "unimplemented feature (was dormant/unregistered before PR #48); un-ignore when implemented"]
 #[test]
 fn test_sv_38_resolver_dispatch() {
     run_positive_compliance_test("tests_advanced", "38_resolver_dispatch.sv");
 }
-#[ignore = "unimplemented feature (was dormant/unregistered before PR #48); un-ignore when implemented"]
 #[test]
 fn test_sv_39_builtin_nettype_resolution() {
     run_positive_compliance_test("tests_advanced", "39_builtin_nettype_resolution.sv");
 }
-#[ignore = "unimplemented feature (was dormant/unregistered before PR #48); un-ignore when implemented"]
 #[test]
 fn test_sv_40_struct_nettype_resolution() {
     run_positive_compliance_test("tests_advanced", "40_struct_nettype_resolution.sv");
+}
+#[test]
+fn test_sv_44_eenet_rnm() {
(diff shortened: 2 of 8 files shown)


vrajeshprakhya:
## Prior art: this overlaps xezim#97

I opened this without searching for existing work first — my mistake. [#97](https://github.com/aionhw/xezim/pull/97) by @opensource-elearning predates it by two weeks and covers the same ground: the simulator-side spread of struct-typed nettype resolver results into member nets, plus a resolver-dispatch regression test. Credit for getting there first belongs to #97.

Leaving this open rather than closing it because #97 and its paired core PR have been stalled since Aug 12, and because the review on #97 and on [xezim-core#18](https://github.com/aionhw/xezim-core/pull/18) named things this branch handles. Happy to close it or fold any single piece into #97 instead — whichever is least work for you.

## Against the reviews

**Addressed here**

- **B3 on core#18 — "no struct nettype test despite the file's own doc comment citing one; the only test declares `typedef logic [7:0] byte_t`, a packed scalar."** This branch un-ignores compliance tests 37–40 and adds `44_eenet_rnm.sv`: a struct of three `real` fields (V/I/R) resolved field-by-field by Millman's theorem across 2–4 simultaneous drivers, plus KCL current summing on a rail. Reference values are hand-derived circuit results, so a resolver that is skipped, folded, or handed a truncated driver set cannot pass.
- **B3, second half — "no changing-driver test; both drivers are constants."** Test 38 drives its nets from variables and steps them across five time steps, so re-evaluation on driver change is exercised.
- **B1 on #97 — `Cargo.toml` repointed at `/home/temp/wt-core18-rebase`.** Not present here; this branch doesn't touch `Cargo.toml` at all. The core dependency stays on the committed pin, and local co-development used the repo's own `scripts/use-local-core.sh` (which writes a git-ignored `.cargo/config.toml`) rather than editing the manifest.

## Something worth flagging about tests 37–40

Those four tests were added alongside 31 but were **never runnable**: each includes `"svtest_defs.svh"` while the harness lives at `"../common/svtest_defs.svh"`, so all four failed in the preprocessor before reaching simulation. Being `#[ignore]`d meant nothing surfaced it.

With the include fixed, **37 (z-skip) and 39 (per-NetType fold) pass against `main` with no code change at all** — built-in net resolution is already implemented and those two were stale-ignored. Their in-file comments predicting `wand [0,1] -> 1  WRONG` are outdated; I left them alone to keep the diff minimal, happy to update them if you'd prefer.

So only 38 and 40 were real gaps, and both were `TEST_FAIL count=7` before the paired core change.

## Ordering

This is a draft because it can't build until the core PR lands — it needs resolver dispatch that isn't in the pinned `rev`. Once [xezim-core#27](https://github.com/aionhw/xezim-core/pull/27) (or #18, if you take that route instead) merges, I'll bump both `rev` fields to the merged SHA and mark this ready, so it goes green in one step rather than showing a red run now.

Two further test branches are prepared but deliberately not proposed yet, since each depends on core work that isn't merged: a package-scope nettype test, and a hierarchy test covering nettype nets driven from several module instances through ports (which fails all five checks without the fix, while the single-module test still passes — so it isolates hierarchy behaviour rather than re-testing what 44 covers).

## Verification

`cargo test --no-fail-fast` and `cargo test --features jit --no-fail-fast` both green against the paired core branch. The simulator commit here was also verified in isolation against *released* core, so it stands on its own.

aionhw (approved):
Reviewed as a pair with core #27 (path-patched worktrees, both branches) and validated independently: compliance 37–40 + EEnet all pass, full suite green with `--features jit` (13 targets, 0 failures), and both branches merge onto current main with zero conflicts.

The two `is_real` projection fixes are correct and well-isolated — `struct_leaf_layout` had already computed the flag both times; binding it instead of `_` is the whole fix. The EEnet test is the best kind of regression: hand-derived circuit values across mixed driver natures mean a skipped/folded/truncated resolver can't sneak through. Good catch on 37/39 being stale-ignored path bugs rather than missing features.

One logistics note for the un-draft: main's core pin moved to `7e2d157` yesterday (a `--dump-merged-sv` API), so when core #27 merges, bump both `rev` fields to the #27 **merge** SHA (which will contain 7e2d157's parent chain once rebased/merged on current core main — #27 is conflict-free against it).

Approving; ready once the pin bump lands after the core merge.

aionhw:
Core #27 is merged — the merge SHA for the `rev` bump is `9377b5a4f360e2249be1d12003a083209a4d6d1b` (it contains `7e2d157`, which current xezim main pins, so bumping both fields to it is safe). Marked ready; happy to merge once the pin commit is pushed.

aionhw:
Landed on main as d012631 + 3a42542 (cherry-picked, your authorship preserved). No pin bump was needed in the end: main's core pin had already moved past xezim-core#27 while fixing #121, so your branch applied cleanly as-is.

Verified after landing: all five resolution-tier tests run and pass (37, 38, 39, 40, 44_eenet_rnm), full suite green across all targets, and the two `is_real` projection fixes confirmed against the issue reproducers — `mk().f1` now reads 1.5 instead of 4609434218613702656. Thanks for the thorough writeup and the EEnet regression; the hand-derived circuit values made it easy to trust.

Issue #60, opened 2026-07-30
Macro definition ending with `\` on last line of file produces duplicate body and literal backslash


overcookedlobster:
Repro

Create `defs.svh` (ensure the file has NO trailing newline after the `\`):

`define FOO(arg) \
  {1, arg} \

Then `top.sv`:

`include "defs.svh"
module top;
  wire [1:0] x;
  assign x = `FOO(1'b0);
endmodule

# Create defs.svh without trailing newline
python3 -c "
data = '\x60define FOO(arg) \\\n  {1, arg} \\\n'
with open('defs.svh', 'wb') as f: f.write(data.encode())
"
xezim --compile -I . -s top top.sv

 Error

top.sv] 6:15: error: expected Semicolon, found LBrace '{'

Preprocessed expansion shows:

assign x =  
  {1, 1'b0}   {1, 1'b0} \;

The macro body `{1, arg}` appears **twice** and a literal `\` appears before the `;`.

assign x = {1, 1'b0};

aionhw:
Fixed in xezim-core `944e3c8`.

**Root cause.** The `\`define` backslash-continuation join loop consumed the text before the backslash, then called `lines.next()` to fetch the continuation line. When the source ended there (your `defs.svh` with no line after the final `\`), `next()` returned `None` and control fell *out* of the `if let` into the unconditional "no continuation here" tail — which appended the same text a **second** time, backslash included. Hence `{1, arg}   {1, arg} \`.

It now stops at the end of the source instead: the text before the backslash has already been taken, and there is nothing to splice onto it.

Your reproducer now preprocesses to `assign x = {1, 1'b0};`, and matches a reference simulator byte-for-byte. Regression tests in `tests/preprocessor_github_issues.rs` cover both the two-source form and a real `\`include` file with no trailing newline, plus an ordinary multi-line `\`define` to guard the common case.

Thanks for the precise reproducer — the "no trailing newline" detail is what made it reproducible.

PR #122, opened 2026-08-22
sim: clockgen must apply 4-state ~ (X never toggles)

Thomasb81:
## Summary

The built-in clock-generator fast path (`always #N clk = ~clk` specialized to an O(1) toggle, in `fire_clock_generators`) turned **any non-`One` bit — including `X`/`Z` on an uninitialized 4-state `logic` — into `1`**, synthesising a 0→1→0 clock out of a signal that, per LRM §11.5.1, stays `X` forever and never produces an edge.

An uninitialized `logic` (no `= init`) is `X` at t=0; driving it with `always #N clk = ~clk` must keep it `X` (no posedge, no edge at all). xezim invented 6 synthetic posedges.

## Fix

`fire_clock_generators` now applies the real 4-state NOT (0→1, 1→0, X→X, Z→X) and **retires the generator** when the signal is currently unknown, so an `X` clock idles instead of clocking. Genuine `clk = 0` self-toggles still toggle byte-identically to the reference (6 posedges over `#60`).

## Verification

- **Standalone pure-SystemVerilog self-test** `tests/clockgen_x_clock_stays_x.sv` (driven via `tests/scheduling/clockgen_x_clock_stays_x.rs` through the CLI, same pattern as `reg3456_pure_sv`):
  - X-start (`logic clkX; always #5 clkX = ~clkX`) → **`TAG_PASS_clock_stays_x`**
  - explicit-0 (`logic clk0 = 0; always #5 clk0 = ~clk0`) → **`TAG_PASS_explicit_toggles`**
- **Pre-fix** the same `.sv` reports `TAG_FAIL_clock_active peX=6` (the bug, demonstrated live).
- Output **byte-identical to the reference simulator** on the fixed binary for both cases.
- Full `--test scheduling` group: 324 passed, 0 failed (the single `phase_jump_static_latch` failure is pre-existing and unrelated — it reproduces identically on the unmodified tree and exercises no clock generators).

### DIFF
--- a/tests/clockgen_x_clock_stays_x.sv
+++ b/tests/clockgen_x_clock_stays_x.sv
@@ -1,0 +1,41 @@
+`timescale 1ns/1ns
+// Pure-SystemVerilog reproduction of the clockgen 4-state `~` bug.
+//
+// An uninitialized 4-state `logic` is X at :0. Per LRM §6.8, ~1'bx == 1'bx,
+// so `always #5 clk = ~clk;` must NEVER toggle an X-start clock (no posedge,
+// no edge at all). xezim's clock-generator fast path used to treat any
+// non-One bit (X/Z included) as 0 and flip it to 1, synthesising a 0->1->0
+// clock out of `clkX` and firing 6 posedges. The explicit `clk0 = 0` clock
+// must still toggle normally (6 posedges over #60).
+//
+// Run: xezim --simulate -s top tests/clockgen_x_clock_stays_x.sv
+// Expect (both the reference simulator and fixed xezim): BOTH tags:
+//   TAG_PASS_clock_stays_x      xezim pre-fix: TAG_FAIL_clock_active pe=6
+//   TAG_PASS_explicit_toggles   both sims agree pe=6
+module top;
+  logic clkX;                     // 4-state, no init -> X at :0
+  always #5 clkX = ~clkX;         // ~X = X -> stays X, never edges
+
+  logic clk0 = 0;                 // explicit init -> genuine clock
+  always #5 clk0 = ~clk0;         // 0->1->0, posedges at 5,15,25,35,45,55
+
+  int unsigned peX = 0;
+  int unsigned pe0 = 0;
+
+  initial forever @(posedge clkX) peX++;
+  initial forever @(posedge clk0) pe0++;
+
+  initial begin
+    $display("CLKX t=%0t clkX=%b", $time, clkX); // report start of X clock
+    #60;                    // posedges on clk0 at 5,15,25,35,45,55 -> 6
+    if (peX == 0)
+      $display("TAG_PASS_clock_stays_x");
+    else
+      $display("TAG_FAIL_clock_active peX=%0d", peX);
+    if (pe0 == 6)
+      $display("TAG_PASS_explicit_toggles");
+    else
+      $display("TAG_FAIL_explicit pe0=%0d", pe0);
+    $finish;
+  end
+endmodule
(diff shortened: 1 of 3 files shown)


aionhw (changes requested):
Thanks — the diagnosis is right, the 4-state NOT table (0→1, 1→0, X→X, Z→X) is exactly what §11.5.1 requires, and the self-checking test with the pre-fix failure banner is a great pattern. Two of the three pieces are mergeable as-is; the **generator retirement is a regression** and needs one small change.

## The retirement kills legitimately late-started clocks

`try_extract_clock_gen` **consumes** the always block (`clock_generators.push(cg); return None;`) — there is no "real (correct) AST/bytecode always block" left to fall back to, so the comment above the retire is describing machinery that doesn't exist. Once `next_toggle_time = u64::MAX`, that clock can never run again — but an X clock is not "unknown forever": the standard *start-the-clock-after-reset* idiom seeds it from the testbench:

```systemverilog
logic clk;                      // X at t=0
always #5 clk = ~clk;
initial begin #20; clk = 0; end // TB starts the clock at t=20
```

Reference behavior: X (no edges) until t=20, first posedge at t=25, then every 10ns → 8 posedges in `#100`. With this branch built and run: **pe = 0 — the clock is permanently dead**, and any TB waiting on `@(posedge clk)` hangs to timeout. (Current main's bug gives pe=10 — wrong, it invents edges before the seed — but at least the clock runs; the retire trades a wrong clock for a dead one.)

## The Z→X first transition is dropped

`continue`-ing before the write also skips the one *genuine* value change an unknown clock makes: with `initial clk = 1'bz;`, the reference performs `clk = ~1'bz = 1'bx` at t=5 — no posedge/negedge, but a real value change (`@(clk)` anyedge waiters, VCD). On this branch the signal stays `z` forever (measured: `changes=1, clk=z` at t=40; reference transitions to X at t=5).

## Suggested change

Keep the 4-state match; replace the retire with a change-gated write plus normal re-arm:

```rust
let new_val = match cur {
    LogicBit::Zero => Value::from_u64(1, w),
    LogicBit::One => Value::from_u64(0, w),
    _ => Value::all_x(w), // ~X = X ; ~Z = X
};
// Unknown clock: ~X == X is a no-op, but ~Z == X is a REAL value change
// (anyedge waiters, VCD), so write only when the value differs — and keep
// re-arming: a TB that seeds 0/1 later ("start clock after reset") must
// revive this clock. One compare per half-period is noise.
if !cur.is_known() {
    if self.signal_table[cg.signal_id] != new_val {
        write_sig!(self, cg.signal_id, new_val);
        if !self.dirty_signals[cg.signal_id] {
            self.dirty_signals[cg.signal_id] = true;
            self.dirty_list.push(cg.signal_id);
        }
        self.dirty_any = true;
        self.table_modified = true;
    }
    cg.next_toggle_time += cg.half_period;
    continue;
}
```

(Your `advance_clock_generators` barrier reasoning was checked and is fine — `u64::MAX` neither spins nor overflows there — it's just not needed once the generator keeps re-arming.)

Could you also extend the test with the late-seed case above (expect 8 posedges) so the revive path stays pinned? With those two changes this is good to merge. Nit: `tests/clockgen_x_clock_stays_x.sv` is missing a trailing newline.

aionhw:
Landed on main as 9fd6a4c with the two review items folded in: unknown clocks now re-arm instead of retiring (change-gated write, so a Z seed still makes its one ~Z == X transition), and the test grew the late-seed (revives, 4 posedges) and Z-goes-X cases. Your commit and authorship are preserved — thanks for the find and the clean repro. Closing since the fix is merged.

Issue #123, opened 2026-08-23
Stack overflow on a uvm_config_db#(virtual iface) set-then-get round trip


vrajeshprakhya:
# Stack overflow on a `uvm_config_db#(virtual iface)` set→get round trip

Passing a virtual interface through `uvm_config_db` — the mechanism essentially every UVM testbench uses to reach its DUT — aborts the simulator:

```
thread 'xezim-sim' (254063) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

`XEZIM_STACK_MB` defaults to **1024**, so this is a 1 GB stack. Raising it further does not help, which points at unbounded recursion rather than depth.

## Reproducer

```systemverilog
import uvm_pkg::*;
`include "uvm_macros.svh"

interface rnm_if;
  int rload;
endinterface

class bare_test extends uvm_test;
  `uvm_component_utils(bare_test)
  virtual rnm_if vif;
  function new(string name, uvm_component parent); super.new(name, parent); endfunction
  function void build_phase(uvm_phase phase);
    super.build_phase(phase);
    if (!uvm_config_db#(virtual rnm_if)::get(this, "", "vif", vif))
      `uvm_fatal("NOVIF", "virtual interface not set")
  endfunction
  task run_phase(uvm_phase phase);
    phase.raise_objection(this);
    vif.rload = 42;
    #10;
    `uvm_info("VIF", $sformatf("rload=%0d", vif.rload), UVM_NONE)
    phase.drop_objection(this);
  endtask
endclass

module top;
  rnm_if vif();
  initial begin
    uvm_config_db#(virtual rnm_if)::set(null, "*", "vif", vif);
    run_test("bare_test");
  end
endmodule
```

```
xezim --simulate -s top \
  -I <UVM>/src -D UVM_NO_DPI -D UVM_REPORT_DISABLE_FILE_LINE \
  <UVM>/src/uvm_pkg.sv repro.sv +UVM_TESTNAME=bare_test
```

Expected: `UVM_INFO ... [VIF] rload=42`. Actual: stack overflow, process aborts.

## It is the round trip, not either half

Narrowed by building the case up piece by piece:

| Case | Result |
|---|---|
| Bare UVM test, no interface | ok |
| `config_db#(virtual iface)::set` only, never `get` | ok |
| `config_db#(virtual iface)::get` only, never `set` | ok |
| A virtual interface handle with no `config_db` at all | ok |
| **`set` then `get` of the same virtual interface** | **stack overflow** |
[...]

Issue #119, opened 2026-08-22
User-defined nettype resolution functions (§6.6.7) are parsed but never called


vrajeshprakhya:
# §6.6.7 user-defined nettype resolution functions are parsed but never called

## Summary

`nettype T wT with resolver;` parses and the resolver name is stored in
`NettypeDeclaration.resolver`, but nothing ever reads that field. Multiple
drivers on a nettype net are collapsed with a hardcoded bitwise-OR fold
(`xezim-core/src/elaborate.rs`, "User-defined nettype driver resolution"), whose
own comment is explicit about it:

> This approximates the common `resolve_or` resolver; other resolvers are not
> modeled, so last-driver-wins behavior applies via the final `|` fold.

The consequence is silent wrong values rather than an error: an OR resolver
passes by coincidence, while AND/XOR/custom resolvers, and every real- or
struct-valued resolver, produce incorrect results with no diagnostic.

## Reproducer

```systemverilog
function automatic logic my_and(logic drivers []);
  automatic logic acc = 1'b1;
  foreach (drivers[i]) acc &= drivers[i];
  return acc;
endfunction

nettype logic and_net_t with my_and;

module top;
  and_net_t n;
  assign n = 1'b0;
  assign n = 1'b1;
  initial begin #0; $display("n=%b (want 0)", n); end   // prints n=1
endmodule
```

A real-typed nettype is wrong in a second way — the net loses `is_real` and
reads back the raw f64 bit pattern:

```systemverilog
function automatic real rsum(input real d[]);
  rsum = 0.0; foreach (d[i]) rsum += d[i];
endfunction
nettype real realnet with rsum;

module top;
  realnet n; assign n = 3.5;
  initial begin #1; $display("n=%f", n); end  // 4615063718147915776.0, want 3.5
endmodule
```

## The repo already has tests for this

`tests/sv_compliance/tests_advanced/` contains four resolution-tier tests added
alongside 31 — 37 (z-skip), 38 (resolver dispatch), 39 (per-NetType fold), 40
(struct-typed nettype, the LRM's `Tsum` example). All four are `#[ignore]`d in
`tests/misc/sv_compliance_runner.rs`.

Two things about them are worth flagging:

1. **They were never runnable.** Each includes `"svtest_defs.svh"` while the
   harness lives at `"../common/svtest_defs.svh"` (31 has the correct path), so
   all four fail in the preprocessor before reaching simulation. The `#[ignore]`
   meant nothing surfaced it.

[...]

Issue #28, opened 2026-07-13
foreach Construct Failure in the Constraints Engine


rharikrishna25:
The following SystemVerilog test case fails with xezim v0.9.0:

<details>

<summary>SV Code with the foreach Construct in the Constraints Set ( sv_constraint_foreach.sv )</summary>

```systemverilog
`ifndef SVTEST_DEFS_SVH
`define SVTEST_DEFS_SVH

`define SVTEST_INIT \
  int failures = 0;

`define SVTEST_CHECK(expr, msg) \
  if (!(expr)) begin \
    failures++; \
    $display("FAIL: %s", msg); \
  end

`define SVTEST_PASSFAIL \
  if (failures == 0) begin \
    $display("TEST_PASS"); \
  end else begin \
    $display("TEST_FAIL count=%0d", failures); \
    $fatal(1); \
  end

`endif

module sv_constraint_foreach ;

  typedef bit [31:0] u32_t ;

  // Unpacked fixed-size array to isolate loop mechanics from dynamic allocation variables
  u32_t test_array[4];

  initial begin
    // Instantiate macro to track test failures
    `SVTEST_INIT

    $display("--- Starting Minimal 'foreach' Constraint Verification ---");

    // Randomize using a highly predictable incremental mathematical constraint
    void'(std::randomize(test_array) with {
      foreach (test_array[i]) {
        test_array[i] == i + 5; 
      }
    });

    // Display results for immediate human debugging visibility
    $display("Resulting Array: %p", test_array);

    // Automated Self-Checking Logic
    // Expected values: test_array[0]=5, test_array[1]=6, test_array[2]=7, test_array[3]=8
    foreach (test_array[i]) begin
      `SVTEST_CHECK(
        (test_array[i] == i + 5), 
        $sformatf("Foreach loop error! test_array[%0d] expected %0d, but got %0d", i, (i + 5), test_array[i])
      )
    end

    // Evaluate global status and conclude test
    `SVTEST_PASSFAIL
  end

endmodule
```
</details>

It appears that the randomization constraints inside the foreach construct are not obeyed.

<details>

<summary>xezim 0.9.0 Execution Log</summary>

```shell
xezim sv_constraint_foreach.sv
```

```text
=== xezim 0.9.0 ===
Max time: 100000
------------------------------
[PHASE] compilation: 0.3ms
[...]

PR #124, opened 2026-08-24
tests: nettype nets across module hierarchy, and in packages (§6.6.7)

vrajeshprakhya:
# tests: nettype nets across module hierarchy, and in packages

Paired with [xezim-core#28](https://github.com/aionhw/xezim-core/pull/28), which implements the hierarchy resolution. Test 46 needs it; test 45 passes on current `main` as-is.

## 46_nettype_hierarchy.sv

Covers the gap you named when closing [#119](https://github.com/aionhw/xezim/issues/119) — nettype nets driven from several module **instances** through ports. Tests 38, 40 and 44 all resolve inside a single module, so none of them reach the topology user-defined nettypes exist for: a shared analog node with drivers and loads contributed by different blocks.

| | | |
|---|---|---|
| **A** | three instances driving one scalar node | `7.0` |
| **B** | a parent-level driver plus an instance driver | `24.0` |
| **C** | struct-valued node shared by a **nested** dut and a parent-level load, real parameters through `#()` overrides | `V=4.615385  R=76.923077  I=0.01` |

**Case C is the one that earns its place.** A nettype net crossing a port boundary is elaborated as a separate signal per port, joined by identity continuous assigns, so resolving them separately runs the resolution function once per port net and again over the results. That compounding is **invisible for an associative resolver** — a plain sum is right either way — so C resolves by Millman's theorem, where resolving twice does not equal resolving once. A and B alone would not catch it.

Reference values are hand-derived circuit results, so a resolver that is skipped, folded, or handed a truncated driver set cannot pass.

**Verified to fail without the fix:** against `main`, all five checks fail while single-module test 44 still passes — so it isolates hierarchy behaviour rather than re-testing what 44 covers.

## 45_package_nettype.sv

A nettype declared in a **package**, referenced both by bare name after a wildcard import and by the qualified `Pkg::name` form, plus two nettypes declared in one package.

The registration this covers landed in xezim-core#27, but nothing upstream exercises it — and a package is the idiomatic place to declare a nettype, since one used on both sides of a port connection has to be named identically in both modules. Before that registration, such a net was not classified as a net at all and a second driver failed with the generic *"Variable 'n' has multiple continuous drivers"*.

This one passes on current `main` with no code change; it is here so the behaviour stays pinned.

## Verification

Full suite **2258 / 0** with the paired core branch, interpreter and `--features jit`. Test 45 also confirmed green against unmodified `main`.

Beyond the compliance suite, the pair makes a UVM testbench able to drive a real-number model through module ports — a `uvm_test` pulling a virtual interface from `config_db`, programming a load resistance, and sampling the resolved node voltage:

```
rload=100.0  -> vnode=2.500000        rload=25.0   -> vnode=1.000000
rload=400.0  -> vnode=4.000000        rload=1000.0 -> vnode=4.545455
UVM_RNM_PASS      UVM_ERROR : 0
```

On current `main` that same testbench reports `vnode=0.000000` for all four. Worth noting it only became runnable at all after your #123 fix — the virtual-interface `config_db` round trip was aborting before that.

### DIFF
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -293,6 +293,10 @@ fn test_sv_47_nettype_coercion() {
     run_positive_compliance_test("tests_advanced", "47_nettype_coercion.sv");
 }
 #[test]
+fn test_sv_48_nettype_inlined_real() {
+    run_positive_compliance_test("tests_advanced", "48_nettype_inlined_real.sv");
+}
+#[test]
 fn test_sv_41_genfor_localparam_idx() {
     run_positive_compliance_test("tests_advanced", "41_genfor_localparam_idx.sv");
 }
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -281,6 +281,14 @@ fn test_sv_44_eenet_rnm() {
     run_positive_compliance_test("tests_advanced", "44_eenet_rnm.sv");
 }
 #[test]
+fn test_sv_45_package_nettype() {
+    run_positive_compliance_test("tests_advanced", "45_package_nettype.sv");
+}
+#[test]
+fn test_sv_46_nettype_hierarchy() {
+    run_positive_compliance_test("tests_advanced", "46_nettype_hierarchy.sv");
+}
+#[test]
 fn test_sv_41_genfor_localparam_idx() {
     run_positive_compliance_test("tests_advanced", "41_genfor_localparam_idx.sv");
 }
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -289,6 +289,10 @@ fn test_sv_46_nettype_hierarchy() {
     run_positive_compliance_test("tests_advanced", "46_nettype_hierarchy.sv");
 }
 #[test]
+fn test_sv_47_nettype_coercion() {
+    run_positive_compliance_test("tests_advanced", "47_nettype_coercion.sv");
+}
+#[test]
 fn test_sv_41_genfor_localparam_idx() {
     run_positive_compliance_test("tests_advanced", "41_genfor_localparam_idx.sv");
 }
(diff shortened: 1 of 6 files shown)


vrajeshprakhya:
Added `47_nettype_coercion.sv`, pinning the §6.6.8 coercion rule in the third commit of [xezim-core#28](https://github.com/aionhw/xezim-core/pull/28) — an implicit net used as a port actual takes the port's user-defined nettype instead of the §6.10 scalar default.

It carries a control half on an explicitly declared net, which turned out to matter more than expected. On a core branch with coercion but *not* the hierarchy fix, that control fails too:

```
core = coercion only          FAIL: implicit net coerced to the port's nettype
                              FAIL: control: explicitly declared nettype net
core = coercion + hierarchy   TEST_PASS
```

Coercion is about a port actual, so a driver has to reach the node through a port — which is the hierarchy case. That is what proved the two changes are inseparable and belong in this pair rather than a third PR, which is what I had originally planned.

Also worth recording here: I did **not** add a test for a bare `interconnect w;` declaration. That form is parse-accepted then discarded, so the name looks undeclared and the new coercion rule gives it the right nettype — a test written that way passes for the wrong reason and pins the side effect rather than the feature. My first draft did exactly that. The interconnect gaps are tracked in [aionhw/xezim#125](https://github.com/aionhw/xezim/issues/125), with a test written against the port form instead, which has no identifier for coercion to act on.

Suite 2259 / 0, interpreter and `--features jit`.

vrajeshprakhya:
Converting this to a draft — the CI failure here is the cross-repo pairing, not a defect in the branch, and it cannot go green until [xezim-core#28](https://github.com/aionhw/xezim-core/pull/28) merges.

This workflow resolves the dependency at `refs/heads/main`:

```yaml
- name: Resolve xezim-core HEAD
  run: echo "CORE_SHA=$(git ls-remote https://github.com/aionhw/xezim-core.git refs/heads/main | cut -f1)" >> "$GITHUB_ENV"
```

so the tests here build against core `main`, which does not yet have the hierarchy resolution or the §6.6.8 coercion they exercise. Reproduced locally against core `main` to confirm that is the whole story rather than assuming it:

```
44_eenet_rnm         TEST_PASS
45_package_nettype   TEST_PASS
46_nettype_hierarchy TEST_FAIL count=5
47_nettype_coercion  TEST_FAIL count=2
```

44 and 45 pass on `main` as expected — 45 pins the package registration from #27, which nothing upstream exercised. 46 and 47 are the two that need #28.

That failure is in fact the evidence those tests are worth having: 46 fails all five checks while single-module 44 still passes, so it isolates hierarchy behaviour rather than re-testing what 44 covers, and 47's control half failing is what showed coercion and hierarchy are inseparable.

I should have opened this as a draft from the start, the way #120 was — same pairing, same reason. Marking it draft now; I will bump it to ready once #28 lands. Nothing to review here in the meantime beyond the test content itself.

This is also a concrete instance of the friction @opensource-elearning raised on #97: a test that depends on a paired core change cannot be green on CI until that change merges, so every such contribution spends a spell red for reasons unrelated to its own correctness.

Core-side CI on #28 was a genuine failure of mine (an unfenced doctest); that is fixed and pushed.

aionhw (approved):
Approve — trial-merged with core#28 onto current main, built, ran: the 7 nettype tests pass and the full suite is green. Merge together with core#28 (plus the usual pin bump).

Issue #126, opened 2026-08-24
Mutable static fixed-size array property in a class has no storage (reads x, not shared)


chenbhil:
## Summary

A fixed-size `static` array property in a class  (`static int S[3]`) is elaborated with no array storage. It collapses to a single scalar shared cell, so:

- element reads return `x` and writes are lost,
- the cross-instance sharing that `static` exists for never happens,
- `foreach` binds a single garbage element,
- `$size(S)` (bare, `this`-scoped) returns `32` which is the collapsed scalar's bit-width, not `3`.

## Repro

```systemverilog
module top;
  class C;
    static int S[3];   // BUG: fixed-size static array — no shared storage
    int        M[3];   // control: non-static member array — stores correctly

    function void report_size();
      $display("$size(S) = %0d   (expect 3)", $size(S)); // bare -> isolates static bug
      $display("$size(M) = %0d   (expect 3)", $size(M)); // control -> 3
    endfunction
  endclass

  initial begin
    C a = new();
    C b = new();

    a.S[0] = 10; a.S[1] = 20; a.S[2] = 30;
    a.M[0] = 1;  a.M[1] = 2;  a.M[2] = 3;

    $display("a.S = %0d %0d %0d   (expect 10 20 30)", a.S[0], a.S[1], a.S[2]);
    $display("a.M = %0d %0d %0d   (expect 1 2 3)",    a.M[0], a.M[1], a.M[2]);
    $display("b.S = %0d %0d %0d   (expect 10 20 30, shared via static)",
             b.S[0], b.S[1], b.S[2]);
    a.report_size();

    $write("foreach a.S: ");
    foreach (a.S[i]) $write("%0d ", a.S[i]);
    $display("  (expect 10 20 30)");
  end
endmodule
```

## Actual vs expected (latest `main`, core `12467b3`)

| expr | actual | expected |
|---|---|---|
| `a.S[i]` | `x x x` | `10 20 30` |
| `b.S[i]` (after `a` writes) | `x x x` | `10 20 30` (shared, §8.9) |
| `$size(S)` (bare, in a method) | `32` | `3` |
| `foreach (a.S[i])` | one garbage element | `10 20 30` |

A non-static `int M[3]` in the same class stores and reads back correctly (`a.M = 1 2 3`, bare `$size(M) = 3`), isolating this to the `static` path.

[...]

Issue #29, opened 2026-07-13
explicit typecast constraint failure in the Constraints engine


rharikrishna25:
The following SystemVerilog test case fails with xezim v0.9.0:

<details>

<summary>SV Code with the explicit typecast Construct in the Constraints Set ( sv_constraint_explicit_typecast.sv )</summary>

```systemverilog

`ifndef SVTEST_DEFS_SVH
`define SVTEST_DEFS_SVH

`define SVTEST_INIT \
    int failures = 0;

`define SVTEST_CHECK(expr, msg) \
    if (!(expr)) begin \
        failures++; \
        $display("FAIL: %s", msg); \
    end

`define SVTEST_PASSFAIL \
    if (failures == 0) begin \
        $display("TEST_PASS"); \
    end else begin \
        $display("TEST_FAIL count=%0d", failures); \
        $fatal(1); \
    end

`endif

module sv_constraint_explicit_typecast;
    `SVTEST_INIT

    logic [7:0]  A; // 8-bit variable
    logic [15:0] B; // 16-bit random variable

    initial begin
        A = 8'hFF; // 255
        
        // According to IEEE Std 1800-2017 Section 11.6.1 & 11.8.2:
        // Inside the 32'(...) explicit static cast context, the sub-expression 
        // operands (A and B) should be context-determined and sized to 32 bits 
        // prior to performing the multiplication. 
        // 255 * B < 1000 can easily be solved if handled in 32-bit space.
        // If truncated to 16-bit intermediate calculation due to solver error, 
        // the constraint solver may fail or find invalid wrapped values.
        
        if (!std::randomize(B) with {
            32'(A * B) < 32'd1000;
            B > 0;
        }) begin
            `SVTEST_CHECK(1'b0, "randomize failed due to constraint solver size-casting breakdown")
        end else begin
            // Double check standard mathematical validity of the solver choice
            `SVTEST_CHECK((32'(A) * 32'(B)) < 32'd1000, "Solver bypassed the 32-bit cast expansion requirement")
        end

        `SVTEST_PASSFAIL
    end
endmodule

```

</details>

The constraint solver fails to follow standard expression bit length expansion rules outlined in IEEE Std 1800-2017 Section 11.6.1 ("Rules for expression bit lengths") and Section 11.8.2 ("Steps for evaluating an expression") when an explicit size cast (e.g., 32'(expr)) wrappers an arithmetic operation inside an inline with {} constraint.

<details>

[...]

PR #127, opened 2026-08-24
Register fixed-size static class arrays under a shared name

chenbhil:
Runtime half of #126; the elaboration half is aionhw/xezim-core#29, which carries the full analysis. Requires the core pin bumped to that merge.

> **Draft until aionhw/xezim-core#29 merges.** This branch intentionally does
> not build from a bare clone yet: it consumes `static_fixed_arrays`, a field
> that only exists in that core PR, while `Cargo.toml` still pins `9eb0e40`.
> CI will fail on the stale pin — expected, not a broken branch. Once core
> merges I will bump both `rev = ...` fields to the merged SHA and amend it
> into this commit, matching the `Pin core <sha>; <dependent change>` convention
> in the history, so every commit stays independently buildable.
>
> It is opened now rather than after, because core#29 in isolation adds a field
> no code in that repo reads — the design question it asks (a separate bounds-
> carrying list, instead of `static_collections` or `array_properties`) is only
> answerable against this consumer, and all the tests and Verilator parity
> evidence live here.

Static members share one store per class (§8.9). Fixed-size static arrays now register **globally under their bare name** at startup rather than per-instance as `<handle>#name` — the same treatment the static queue and associative collections already get, right above this code. Bounds are the declared ones rather than the 0..63 queue buffer, and every element is defaulted so an index read or `foreach` sees a populated range instead of `x`.

Three call sites ask "is this property a static collection?" to choose between the collection paths and the per-instance packed-select / struct-leaf machinery. All three now count these arrays too:

- a fixed static array indexes its unpacked dimension first (§7.4.2), exactly like the dynamic collections, so it must not reach the packed bit-select path — that misrouting is what made `S[0]` a bit select in the first place;
- its `[i]` element leaves are not reachable per-instance, so the struct-member machinery must not claim them either.

The typedef-dims classifier (`classify_typedef_collection_properties`) gains the same two dimension arms as `elaborate_class`, so `typedef int arr_t[3]; static arr_t S;` reaches the same storage. This is a genuinely separate registration path: a typedef-typed property carries no declarator dims and recovers its shape from the typedef table at simulator startup.

### Testing

`tests/classes/static_fixed_array_storage.rs`, two tests:

- **the issue repro** — element reads, cross-handle sharing, `$size`, and `foreach`, with a non-static `int M[3]` in the same class as the control that was never broken;
- **declaration forms that reach this path** — `[lo:hi]`, parameter-sized `[N]`, typedef-carried dims, `byte` elements, a static array inherited from a base class, a write from a static method (no `this` in scope), and class-scoped `C::S[i]` access with no handle involved.

Every assertion was first confirmed against Verilator 5.050 as reference.

Full suite: **1887 passed**. The 7 failures are all `dpi_integration_tests` aborting at `cc failed`, which reproduces on a clean checkout too — building the DPI stubs `-shared` on macOS arm64 leaves the VPI symbols (`_vpi_get`, `_vpi_free_object`, …) undefined at link time, so it fails in `cc` before xezim runs. Unrelated to this change; needs `-Wl,-undefined,dynamic_lookup` on that platform.

Found 2 separate bugs, will file an issue later.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -5409,6 +5409,7 @@ impl Simulator {
                     || cd.array_properties.contains_key(pname)
                     || cd.array_nd_properties.contains_key(pname)
                     || cd.static_collections.iter().any(|(n, ..)| n == pname)
+                    || cd.static_fixed_arrays.iter().any(|(n, ..)| n == pname)
                 {
                     continue;
                 }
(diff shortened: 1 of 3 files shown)


aionhw (commented):
Runtime half reviewed together with core#29 (built and tested as a pair on current main — see the two requested fixes over there: 4-state elements default 0 instead of x, and two classes with same-named static arrays share one store). The draft-until-core-merges pin dance is handled per convention. Once core#29 lands with those fixes and the pin bumps, this is good to go.

PR #130, opened 2026-08-24
Fix parameterized static dispatch: no-parens calls, value-param scope, nested type-arg resolution

Thomasb81:
## Summary

This PR fixes the UVM compat test **7443** and a chain of related SystemVerilog semantics bugs it surfaced (six commits stacked). The final commit produces a byte-for-byte match against a commercial reference simulator.

### 1. No-parens parameterless static method invocation (LRM §13.4.1)
A `Class::method` reference **without parentheses** where `method` is a parameterless static function must be *called*, not read as a non-existent static property. Both the flat 2-segment `Ident` form and the `MemberAccess` form (receiver = class type-parameter, e.g. `TYPE::x`) fell through `class_static_get`, which only reads static *properties*, so the call returned the empty property cell.

UVM's `uvm_misc.svh` warning
```systemverilog
{"More than one instance of type '", TYPE::type_name, " found ..."}
```
in 7443 (with `TYPE` bound to `comp_b`) rendered the type name **empty** (`type ' found`) where the reference prints `type 'comp_b found`.

### 2. Value-params from the caller scope (§13.5.1)
A value parameter of an enclosing generic class passed as an actual argument into a method of a *different* parameterized class now resolves in the **caller's** scope, not against the callee's (otherwise it falls back to its declared default). Implemented via a `spec_scope_stack` walked during value-param resolution.

### 3. Supporting chain
- T-bound value-param queues through nested generic method chains
- Bare T-bound unpacked-array members as collection storage
- Register type-param-bound collection members as per-instance arrays

## Validation
- 7443 passes with 0 errors / 0 fatals; the `find_type-multi match` warning renders `type 'comp_b found`, byte-for-byte matching a commercial simulator.
- Two new pure-SV self-tests fail without their respective fixes and pass with them:
  - `t_bound_generic_value_param_argument.rs` (value-param scope)
  - `t_no_parens_static_method_invocation.rs` (no-parens static method call)
- Full suite: 315/317 (2 pre-existing, unrelated `uvm_config_db_tests` failures); 150/150 in the hierarchy group.

---

## Update: nested value-parameter resolution in type arguments

Added `Resolve nested value parameters in type arguments from active spec` (top commit):

An inner `Comp#(N)` type argument carries a SYMBOLIC value parameter `N` at the point its static-method body is evaluated. The parser reconstructs parameterized type text with spaces around `#`/parens (`Comp # ( N )`), but the spec extractor only matched the compact `#(` form, so the nested spec was never parsed and `N` stayed unresolved (fell back to declared default `0`). This split per-specialization singletons/typeids that UVM relies on. The extractor is now whitespace-tolerant.

Procedure check: self-test `t_nested_value_param_in_type_arg` fails without the fix (N2_FAIL) and passes with it (N2_PASS, byte-for-byte matching a commercial simulator). The dependent UVM tests are green on this branch:
- 05components/90Mantis/5978 -> PASS
- 05components/99compat/6533_print_config_matches -> PASS
- 05components/99compat/7443 -> PASS

These three require the earlier value-param-scope commits in this same PR (they do NOT pass on a main-only branch); with the full stacked branch they all pass. Full self-test suite: 315 passed / 2 pre-existing unrelated failures.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -93292,6 +93292,48 @@ impl Simulator {
             }
         }
         self.heap.push(Some(instance));
+        // Per-instance TYPE-PARAM-BOUND collection members
+        // (`#(type T=int) data;` instantiated with T bound to a queue/array
+        // typedef such as `typedef int a_i[];`). These live only in
+        // `cd.properties` generically and never reach `queue_properties` /
+        // `array_properties`, so whole-assignment / size / index / `%p` paths
+        // had no array table entry and treated `data` as a scalar. Register
+        // `<handle>#member` as a dynamic array once this instance's type
+        // bindings (set above) resolve the parameter to a concrete collection
+        // type.
+        for cdef in &classes_to_init {
+            for (prop, sig) in &cdef.properties {
+                if cdef.queue_properties.contains_key(prop)
+                    || cdef.assoc_properties.contains_key(prop)
+                    || cdef.array_properties.contains_key(prop)
+                    || cdef.array_nd_properties.contains_key(prop)
+                    || cdef.static_collections.iter().any(|(n, ..)| n == prop)
+                {
+                    continue;
+                }
+                if !self.prop_bound_collection(handle, &cdef.name, prop) {
+                    continue;
+                }
+                let raw = sig.type_name.as_deref().unwrap_or(prop);
+                let concrete = self
+                    .heap
+                    .get(handle)
+                    .and_then(|o| o.as_ref())
+                    .and_then(|i| i.type_bindings.get(raw).cloned())
+                    .unwrap_or_else(|| raw.to_string());
+                let w = self
+                    .module
+                    .typedef_elem_widths
+                    .get(&concrete)
+                    .copied()
+                    .unwrap_or(sig.width)
+                    .max(1);
+                let scoped = format!("{}#{}", handle, prop);
+                self.module.dynamic_arrays.insert(scoped.clone());
+                self.module.arrays.insert(scoped.clone(), (0, 63, w));
+                self.set_queue_size(&scoped, 0);
+            }
+        }
         // Re-evaluate scalar property initializers against the live parameter
         // table and instance context, before the constructor runs (SV applies
         // member initializers prior to `new`'s body). `elaborate_class`
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -47863,6 +47863,24 @@ impl Simulator {
                                 .signal_name_to_id
                                 .contains_key(self.resolve_hier_name(hier).as_str())
                         {
+                            // `Class::method` with NO parens invokes a
+                            // parameterless STATIC function (LRM §13.4.1: a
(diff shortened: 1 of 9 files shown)


aionhw (commented):
Reviewed by trial-merging onto current main together with core#30, building, and running everything: all 154 hierarchy tests and the full suite (13/13 groups, both feature configs' worth) pass, and the headline no-parens bug is confirmed real on main (`TYPE::type_name` renders empty) and fixed here (`type 'comp_b' found`). Notes before merge:

1. **Hard dependency on core#30.** I rebuilt this branch against core *main*: `percent_m_module_init_scope` fails (prints the declaring module's name). Merge order per repo convention: core#30 first, then this with the `Pin core <sha>` bump amended in — otherwise CI is red on that one test.
2. **`q == q` self-compare shortcut vs x elements.** `compare_dyn_arrays` returns 1 for same-storage operands regardless of contents, but with an `8'hxx` element a commercial reference simulator yields 0 for `q1 == q1` (and 0 for `q1 != q1`). Suggest dropping the shortcut and letting the element loop — which already X-propagates — decide; same storage is then just the trivially-equal walk.
3. **`spec_scope_stack` push/pop is duplicated 9×** with save/restore around each `exec_static_method` call. Any future early return between push and pop leaks a stack entry. All nine are balanced today; a small guard (or closure helper) would keep it that way.
4. **`dyn_cmp_operand_name` robustness**: if a bare name is stale-registered in `module.dynamic_arrays` (the very leak your `resolve_coll` fix documents), a scalar==scalar could reroute through the queue path and compare sizes 0==0. Plain scalars are fine today (verified), but resolving to a plain signal id first as a rejection test would make it robust against that class.
5. Your PR text mentions "2 pre-existing `uvm_config_db_tests` failures" — the full suite is green here on current main + this branch. Which base SHA were those on? If they were environmental (UVM assets), worth a line in the PR body so nobody chases them.

Thomasb81:
> 5\. Your PR text mentions "2 pre-existing `uvm_config_db_tests` failures" — the full suite is green here on current main + this branch. Which base SHA were those on? If they were environmental (UVM assets), worth a line in the PR body so nobody chases them.

This are probably installation artifact issue at my end. Did not dig them but they are still persistent in my working environment.

I propose to ignore this message for the time being.

aionhw:
Merged (`bf26692`), sequenced after core#30 + the pin bump as discussed. Two of the review notes are addressed on top in the same push:

- **`q == q` with x elements** — the same-storage shortcut is gone and an unknown element now makes BOTH `==` and `!=` yield 0, matching the reference simulator's behavior exactly (`SELF_0_0`); known-equal pairs still compare 1/0. Pinned in `tests/collections/queue_eq_unknown_elems.rs`.
- Your env-side note on the `uvm_config_db` failures is recorded here — suite is fully green on main CI-side, so nothing to chase.

Remaining non-blocking notes (spec-stack guard helper, `dyn_cmp_operand_name` signal-id rejection test) are yours to take or leave in a follow-up. Thanks — the byte-for-byte 7443 result was a satisfying one to verify.

PR #131, opened 2026-08-25
Resolve nested value parameters in type arguments from active spec

Thomasb81:
Fixes the UVM `09callbacks/25params`-adjacent value-parameter resolution: a value parameter used in a **nested type-argument position** inside its own parameterized class's static method (`Comp#(N)::get_holder()` referencing `Registry#(Comp#(N))`) was falling back to the parameter's declared default instead of the active specialization's binding.

**Root cause:** the parser reconstructs parameterized type text with spaces around `#`/parens (`Comp # ( N )`), but `extract_spec_from_string` only matched the compact `#(` form, so a nested symbolic value-param spec was never parsed and `N` stayed unresolved → `Registry#(Comp#(0))` instead of `Registry#(Comp#(2))`.

**Fix:** make the spec extractor whitespace-tolerant (locate `#`, allow optional spaces before `(`, trim the base, return the raw sig for the caller's usual `normalize_spec_ws` canonicalization).

**Validation:** `t_nested_value_param_in_type_arg` fails without the fix (`N2_FAIL`) and passes with it, byte-for-byte matching a commercial simulator (`N2_PASS`). Full self-test suite: 315 passed / 2 pre-existing unrelated failures.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -93292,6 +93292,48 @@ impl Simulator {
             }
         }
         self.heap.push(Some(instance));
+        // Per-instance TYPE-PARAM-BOUND collection members
+        // (`#(type T=int) data;` instantiated with T bound to a queue/array
+        // typedef such as `typedef int a_i[];`). These live only in
+        // `cd.properties` generically and never reach `queue_properties` /
+        // `array_properties`, so whole-assignment / size / index / `%p` paths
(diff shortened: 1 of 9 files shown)


Thomasb81:
Fold the nested-value-param fix into PR #130 (this PR's branch is a prerequisite stack subset of #130's). Closing to keep a single coherent PR carrying all the related parameterized-static/value-param fixes.

PR #134, opened 2026-08-25
lint: add check_param_class_scope (§8.25) and check_untyped_port_net_assign (§23.2.2.3 + §6.5)

opensource-elearning:
This PR implements two additive LRM-conformance checks in the `should_fail` second-pass lint (`src/should_fail_lint.rs`):

## 1. Reject unadorned parameterized-class scope resolution (§8.25)
- **sv-tests:** `chapter-8/8.25.1--parametrized_class_invalid_scope_resolution.sv`
- **LRM 1800-2017 §8.25:** "For a parameterized class C, the default specialization is C#(). Other than as the prefix of the scope resolution operator, use of the unadorned name of a parameterized class shall denote the default specialization."
- An unadorned parameterized class used as a `::` prefix (`par_cls::b`) does NOT denote the default specialization and is illegal; a legal form must specialize explicitly (`par_cls#()::b`, `par_cls#(15)::b`).
- The AST distinguishes the two: the legal form's `MemberAccess` base is a `Specialization`, the illegal form's base is a bare single-segment `Ident`. The lint walks every statement/continuous-assign expression and fires only when that base names a class whose HEADER `#(...)` list is non-empty (`param_order` / `type_param_names`). Body-only localparams do NOT parameterize a class, so UVM-style static-call initializers are untouched.

## 2. Reject procedural assignment to untyped output/inout ports (§23.2.2.3, §6.5)
- **sv-tests:** `chapter-14/14.3--clocking-block-signals-error.sv`
- **LRM §23.2.2.3:** An ANSI output/inout port with an omitted or implicit data type defaults to a NET of default net type; procedural assignment (`=`/`<=`) to such a net is illegal (§6.5). A port explicitly typed with `logic`/`var` or completed by a body variable declaration is a variable and legal.
- Extended `check_proc_net_assign` to collect untyped output/inout ANSI ports as implicit nets, excluding ports with `var_kw`, explicit `data_type`, or a body `DataDeclaration` completion (mirrors the existing port-completion logic in elaborate.rs).

## Validation
- All 2,249 tests pass (0 failures)
- Build compiles clean
- sv-tests chapter suite: **823/830 pass** (7 remaining are SVA/constructs not yet supported)
- Both new checks are purely additive — they only add diagnostics after successful elaboration, never change compile/elaborate behavior.

### DIFF
--- a/src/should_fail_lint.rs
+++ b/src/should_fail_lint.rs
@@ -12,14 +12,16 @@
 //! and the whole pass is validated to hold the static-suite baseline (1005).
 
 use xezim_core::ast::decl::{
-    ClassDeclaration, ClassItem, ClassMethodKind, ClassQualifier, ModuleItem, TypedefDeclaration,
+    ClassDeclaration, ClassItem, ClassMethodKind, ClassQualifier, ModuleItem,
+    TypedefDeclaration,
 };
 use xezim_core::ast::expr::{
     AssignmentPatternItem, Expression, ExprKind, NumberBase, NumberLiteral, RangeKind,
 };
+use xezim_core::ast::module::PortList;
 use xezim_core::ast::stmt::{Statement, StatementKind, VarDeclarator};
 use xezim_core::ast::types::{
-    DataType, EnumType, IntegerAtomType, PackedDimension, Signing, UnpackedDimension,
+    DataType, EnumType, IntegerAtomType, PackedDimension, PortDirection, Signing, UnpackedDimension,
 };
 use xezim_core::elaborate::ElaboratedModule;
 use xezim_core::SourceDefinition;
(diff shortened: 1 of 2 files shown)


aionhw (commented):
Update after `a918829` landed on main: the one blocking finding is now moot — the parser propagates `var` across ANSI port-list commas (`output var a, b` makes `b` a variable), so `check_untyped_port_net_assign` no longer has that false-positive window; there's a regression test on main (`tests/misc/interconnect_and_var_ports.rs::var_kw_continues_across_port_comma`). Rebase onto current main and this is reviewable as-is. Remaining earlier notes are non-blocking: the §8.25 walk skips `While`/`DoWhile`/`Repeat` bodies and `Return` expressions (under-reporting only), and package-scoped parameterized classes aren't collected (same direction). Please also close #104 in favor of this one — same content, and it conflicts on `msrv.yml`.

opensource-elearning:
@aionhw Rebased onto latest main with fixes applied:

- Added missing statement handlers in `check_param_class_scope` for `While`, `DoWhile`, `Repeat`, and `Return` expressions
- Collect package-scoped parameterized classes from `elab.classes` (using `param_order` / `type_param_names` same as `class_is_parameterized`)
- Updated `str_param_spec.sv` test to use LRM-conformant `Wrapper#()::type_name()` form
- Removed unused imports

All 2,058 tests pass.

aionhw:
Heads-up: this branch is stacked on #94 (`fix: clamp --threads...` is its first commit), and #94 was just closed without merge — `--threads` is not working as expected and will be removed rather than clamped, so that commit must not land.

The lint work itself (the §8.25 and §23.2.2.3/§6.5 checks) remains approved from the earlier review round. Please rebase onto current `main` dropping the threads commit (and its `tests/perf/threads_clamp.rs` / `tests/fixtures/tiny.sv` companions), and this can merge on its own.

opensource-elearning:
Rebased onto current main (0ef7e37) and dropped the #94 threads commit (`dc71219`) along with its `tests/perf/threads_clamp.rs` and `tests/fixtures/tiny.sv` companions. The PR is now a single clean commit over `main`, touching only `src/should_fail_lint.rs` and `tests/lrm_9_value_param/str_param_spec.sv`. #104 is already closed.

PR #136, opened 2026-08-26
tests: register the negative tests that never ran (neg09-neg11)

vrajeshprakhya:
Paired with **[xezim-core#33](https://github.com/aionhw/xezim-core/pull/33)**. Opened as a draft: CI resolves the core dependency at `refs/heads/main`, so `neg10` cannot go green until #33 merges. Marking ready then.

`tests_negative/` has four tests — neg09 through neg12 — that `a79dad3` added but never registered in `sv_compliance_runner.rs`, so **they have never executed**. Registering them turned up that "four tests were never wired up" was hiding three different problems:

| test | what was actually wrong |
|---|---|
| neg09 duplicate port name | nothing — just unregistered |
| neg11 missing named port | the **harness** did not recognise a correct diagnostic |
| neg10 duplicate enum literal | a real elaboration gap — fixed in core#33 |
| neg12 duplicate case item | the **test's expectation** looks wrong |

## neg11 was a harness bug, not a tool bug

xezim rejects the construct properly — exit 1, precise, §-cited:

```
error: port 'missing' is not a port of module 'neg11_child' (LRM 1800-2017 §23.3.2)
```

But `run_negative_compliance_test` accepted only `": error:"`, `"Parse errors"`, or `"Simulation error"`. The first matches a *parse* diagnostic's source prefix (`[file] 2:15: error: ...`); the third matches elaboration errors raised inside a scope. Neither matches the **bare** `error:` an elaboration check emits when it runs before a scope is established — so a test whose construct xezim correctly rejects still failed the harness.

Widened to also accept a line **starting** with `error:`, matched at line start so the `"0 errors"` summary line cannot satisfy it.

## neg12 stays unregistered, deliberately

It expects `compile_fail` for a repeated `unique case` item. §12.5.3 makes unique/priority violations a **runtime** report, not a compile error, so registering it would pin an expectation the LRM does not support. A statically duplicated constant item is a reasonable *lint* — but the test's own expectation is what needs settling, not the tool. Left in place, unregistered, with the reasoning recorded in the commit message so it is not rediscovered from scratch.

## Result

Negative tests now run **11**, up from the 8 that were actually running before this branch.

## Verification

Against core#33:

- negative suite: **11 passed, 0 failed**
- full suite: **2303 passed / 0 failed** interpreter, **2307 passed / 0 failed** `--features jit`

### DIFF
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -81,10 +81,21 @@ fn run_negative_compliance_test(filename: &str) {
         filename,
         combined
     );
+    // A parse diagnostic carries a source prefix (`[file] 2:15: error: ...`);
+    // elaboration reports either `Simulation error: ...` or, for a check that
+    // runs before a scope is established, a BARE `error: ...` with no prefix
+    // at all. The bare form was not accepted here, so a test whose construct
+    // xezim correctly rejects still failed the harness -- which is why
+    // neg11_missing_named_port could not be registered. Matched at a line
+    // start so the "0 errors" summary line cannot satisfy it.
+    let has_diagnostic = combined.contains(": error:")
+        || combined.contains("Parse errors")
+        || combined.contains("Simulation error")
+        || combined
+            .lines()
+            .any(|l| l.trim_start().starts_with("error:"));
     assert!(
-        combined.contains(": error:")
-            || combined.contains("Parse errors")
-            || combined.contains("Simulation error"),
+        has_diagnostic,
         "Negative test {} failed without an error diagnostic. Output:\n{}",
         filename,
         combined
@@ -348,4 +359,12 @@ fn test_sv_neg07_bad_clocking_direction() {
 #[test]
 fn test_sv_neg08_bad_constraint_reference() {
     run_negative_compliance_test("neg08_bad_constraint_reference.sv");
+}
+#[test]
+fn test_sv_neg09_duplicate_port_name() {
+    run_negative_compliance_test("neg09_duplicate_port_name.sv");
+}
+#[test]
+fn test_sv_neg11_missing_named_port() {
+    run_negative_compliance_test("neg11_missing_named_port.sv");
 }
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -365,6 +365,10 @@ fn test_sv_neg09_duplicate_port_name() {
     run_negative_compliance_test("neg09_duplicate_port_name.sv");
 }
 #[test]
+fn test_sv_neg10_duplicate_enum_literal() {
+    run_negative_compliance_test("neg10_duplicate_enum_literal.sv");
+}
+#[test]
 fn test_sv_neg11_missing_named_port() {
     run_negative_compliance_test("neg11_missing_named_port.sv");
 }


vrajeshprakhya:
CI is red for the cross-repo pairing, not for a defect in this branch. Recording the specifics so the check does not have to be re-read.

Exactly one test fails, in both modes:

```
thread 'sv_compliance_runner::test_sv_neg10_duplicate_enum_literal' panicked at
tests/misc/sv_compliance_runner.rs:78:5:
Negative test neg10_duplicate_enum_literal.sv unexpectedly succeeded.

test result: FAILED. 808 passed; 1 failed; 6 ignored   (cargo test)
test result: FAILED. 812 passed; 1 failed; 6 ignored   (cargo test --features jit)
```

The workflow resolves the core dependency at `refs/heads/main`, which does not yet have the §6.19 check from [xezim-core#33](https://github.com/aionhw/xezim-core/pull/33). So `neg10` is asserting a rejection that unpatched core does not perform.

Worth noting what the failure text actually says: **"unexpectedly succeeded"** — xezim *accepted* `IDLE` declared twice in one enum. That is the gap #33 closes, reproduced by CI against current main rather than argued for in prose.

Two things this run also settles on their own:

- **neg09 and neg11 pass here**, against unpatched core. They needed no core change — only registration, and (for neg11) widening the harness to recognise a bare `error:` diagnostic it was already emitting correctly. Had all three been bundled behind #33, that would not have been visible.
- The other 808 / 812 tests are unaffected, so the harness change does not loosen anything: no previously-failing negative test started passing.

Against core#33 locally the same suites are **2303 passed / 0 failed** (interpreter) and **2307 passed / 0 failed** (`--features jit`).

Staying a draft until #33 merges; I will mark it ready and confirm the run goes green then.

aionhw (approved):
Reviewed and verified end-to-end against **current** `main` (`987e4bc`) and **current** core (`1d62b1b`) — both repos have moved since this was opened, so I re-ran the pairing rather than trusting the original numbers:

- Merges into current `main` cleanly (no conflict with the `49_specparam_hierarchical` registration that landed in the same file since).
- With core#33 applied to current core: **negative suite 11/11**, compliance runner 67/0, core `cargo test` 48/48, **full xezim suite 2317 passed / 0 failed**.
- Independently confirmed the harness diagnosis before reading the fix: `--no-sim` on neg11 emits a **bare** `error:` line that the old matcher missed, while neg09 emits the prefixed `Simulation error:` form it already accepted — which is exactly why neg09 needed only registration and neg11 needed the widening. The line-start anchor holds: neither the `0 errors, 0 warnings` summary nor a `warning:` line can satisfy `starts_with("error:")`.
- The neg12 call is right, and the reasoning in the body matches my own reading: a repeated `unique case` item is a §12.5.3 **runtime** violation-report condition, not a compile error, so registering it would pin an expectation the LRM does not back. Leaving the file in place with the rationale recorded is the correct move.

The four-way split in the PR body — unregistered / harness bug / real tool gap / wrong expectation — is the most useful part: "the tests were never wired up" was hiding three distinct defects, and un-bundling neg09/neg11 from the core dependency is what made each visible on its own.

**Approve.** Merge order: xezim-core#33 first (verified green against current core above), then mark this ready and let CI confirm — nothing in the branch itself needs changing.

Issue #35, opened 2026-07-14
mixed-sign relational operators failure in the Constraints engine


rharikrishna25:
The following SystemVerilog test case fails with xezim v0.9.3:

<details>

<summary>SV Code with the mixed-sign relational operators in the Constraints Set ( sv_constraint_mixed_sign_operators.sv )</summary>

```systemverilog

`ifndef SVTEST_DEFS_SVH
`define SVTEST_DEFS_SVH

`define SVTEST_INIT \
    int failures = 0;

`define SVTEST_CHECK(expr, msg) \
    if (!(expr)) begin \
        failures++; \
        $display("FAIL: %s", msg); \
    end

`define SVTEST_PASSFAIL \
    if (failures == 0) begin \
        $display("TEST_PASS"); \
    end else begin \
        $display("TEST_FAIL count=%0d", failures); \
        $fatal(1); \
    end

`endif

class sign_test_class;
    rand int rand_signed;            // Signed 32-bit: Max = 2147483647, Min = -2147483648
    rand logic [31:0] rand_unsigned; // Unsigned 32-bit: Max = 4294967295
endclass

module test;
    `SVTEST_INIT

    initial begin
        sign_test_class item = new();
        bit success;

        // ====================================================================
        // CASE 1: Mixed Sign Relational Operator (LRM 11.8.1)
        // Mixing a signed variable with an unsigned literal forces the entire
        // expression to be evaluated as unsigned.
        // ====================================================================
        item.rand_unsigned = 0;
        success = item.randomize() with { 
            rand_unsigned <= 32'hffff_ffff; // Unsigned evaluation context
            rand_unsigned >= 4294967290;
        };
        `SVTEST_CHECK(success && (item.rand_unsigned >= 4294967290), 
                      "Case 1: Failed to evaluate mixed-sign expression as unsigned")

        // ====================================================================
        // CASE 2: Strict Signed Contradiction (LRM 11.8.1)
        // When all operands are signed, signed comparisons apply.
        // rand_signed <= 32'shffff_ffff (-1) AND rand_signed >= 1 is impossible.
        // Must return success = 0 (Failure).
        // ====================================================================
        item.rand_signed = 0;
        success = item.randomize() with { 
            rand_signed <= 32'shffff_ffff; // Signed -1
            rand_signed >= 1;              // Signed 1
        };
        `SVTEST_CHECK(success == 0, "Case 2: Failed to catch signed contradiction range")

        // ====================================================================
[...]

PR #139, opened 2026-08-26
Parameterized static dispatch, type-param/specialization, and collection fixes

Thomasb81:
## Parameterized static dispatch, type-param/specialization, and collection fixes

A set of related SystemVerilog semantics fixes in the simulation runtime, all validated byte-for-byte against a reference simulator where stated.

The commits form one interdependent chain (each builds on the prior, all in `src/compiler/simulator.rs`), so they land together.

Highlights:
- **No-parens parameterless static method invocation** (IEEE 1800-2023 §13.4.1), with value-param scope resolution from the caller and nested type-arg resolution.
- **Per-specialization static collection storage**: external `Class#(spec)::member` access to a static assoc array now keys the same per-spec cell as internal static methods (fixes the `.size()`/element-read mismatch and the reuse-pool pattern).
- **Sibling-class same-named static collections** isolated by qualified key.
- **Type-param-bound collections** registered as per-instance arrays; T-typed value-param queues carried through nested generic method chains.
- **Enum introspection** on type-param-typed locals; local enums default to their first member.
- **Signedness** kept on same-width class-member assignments; unsized based literals at natural width; `%p`/rendering of type-param-typed string and signed-int symbols.
- **$typename** of a class type param and typedef-to-builtin; nested typedef spec args through the enclosing class specialization.
- **Virtual-interface resource keys**: `virtual <iface>#(params)` rendered consistently so `uvm_resource_db#(...)::set` and `::read_by_type` find the same pool cell (depends on xezim-core PR #34).

Adds ~25 self-tests across `tests/collections`, `tests/types`, `tests/hierarchy`, `tests/classes`.

Full self-test suite: 325 passed, no regressions (2 pre-existing config-db scope-matching failures, unrelated).

**Dependencies:** requires xezim-core PR **#34** (virtual-interface type args) and **#35** (unsized based literal width). The `Cargo.toml` core `rev` pin must be bumped to the merged core `main` revision once those land.


aionhw:
**Unblocked** — the dependency this needed is now on `main`.

Both CI checks were failing for a single reason, not anything wrong with the change: it uses `DataType::Interface { type_args }`, the field added by xezim-core#34, which had not landed. Three `E0026` errors, all "variant `DataType::Interface` does not have a field named `type_args`" (`simulator.rs:95161`, `:95818`, `:104903`). CI resolves xezim-core at `refs/heads/main`, so nothing in this PR had to change.

xezim-core#34 is now merged (`1d62b1b`), and `main` pins it (`0d016bc`). Please re-run CI.

I verified this branch end to end before merging the dependency: merged current `main` into it (**no conflicts**), built against core#34, and ran the full suite — **2335 passed, 0 failed**. So this is ready on the merits.

One housekeeping note: this branch still pins the old core rev in `Cargo.toml`. Merging `main` picks up the bumped pin, which per the comment in `Cargo.toml` is what keeps a bare clone building.

Issue #137, opened 2026-08-26
§6.6.7 resolver dispatch never compiles: Expr_Call_ports on a dynamic-array resolver formal dominates RNM runtime


vrajeshprakhya:
Follow-up to #129, which invited this:

> Remaining ceiling from your report: `Expr_Call_ports` on the §6.6.7 resolver dispatch (`sum_currents('{...}')` — an assignment-pattern actual into a dynamic-array formal). That shape is still AST-evaluated; worth its own issue if it dominates on your 4-block DAC model.

It dominates. Measured on current `main` (`39ca183` + core `5e46fda`), release build, x86_64 Linux.

## The 4-block current-summing DAC

```
run 1: 128.41 s      jit run 1: 124.37 s
run 2: 118.32 s      jit run 2: 117.17 s
run 3: 116.77 s
```

```
[PROF] settle_ca=111040.0ms (218305 evals)
[PROF] simulation_loop=116847.0ms
[PHASE] total=117419.0ms
[PROF] insns=189480  edges_fired=0
```

`settle_ca` is **111 s of a 117 s total — about 95%**. Only 189,480 bytecode instructions execute in the entire run, so effectively none of the hot path is compiled. **`--features jit` changes nothing**, which is the expected consequence rather than a separate puzzle: there is no compiled code for it to improve.

```
[COMPILE_FAIL_STATS] continuous assign bytecode failures:
       4 Expr_Call_impure
       1 Expr_Call_ports
```

Correctness is not the issue — this model last measured `rms_error_norm=0.0976` against its ngspice reference. This is purely time.

## Self-contained reproducer

The same shape reduced to four blocks and a passive load, no generated code:

```
[COMPILE_FAIL_STATS]        1 Expr_Call_ports
out_val=0.041272251  isum=0.000082545

wall: 2.92 s
[PROF] settle_ca=2532.2ms (100002 evals)
[PROF] simulation_loop=2972.3ms
[PHASE] total: 2976.6ms
```

**85% in `settle_ca`** for 20k timesteps, in about 3 seconds.

```systemverilog
// Minimal shape of a current-summing RNM: a §6.6.7 user-defined nettype whose
// resolution function takes a DYNAMIC-ARRAY formal, driven by several module
// instances onto one shared node, with the node's own passive load closing the
// loop. This is the structure spice2rnm emits for a current-mode block.
`timescale 1ns/1ps

package isum_pkg;
  // KCL: the node's value is the sum of everything driving it.
[...]

PR #140, opened 2026-08-27
tests: Verilog-AMS  nets (pairs with xezim-core#36)

vrajeshprakhya:
Compliance tests for the `wreal` support in **aionhw/xezim-core#36**.

- `50_wreal_nets.sv` — scalar and port forms; single-driver pass-through, and
  a three-driver sum checked against the arithmetic total rather than any one
  driver, so a "last write wins" resolution would fail it rather than pass by
  luck.
- `neg13_wreal_packed_range.sv` — a packed range on a `wreal` must be
  rejected. A `wreal` is a real; there are no bit positions to select.

### Ordering

xezim pins xezim-core by exact rev, so **core#36 must land and the rev be
bumped before this can pass CI**. Until then the two new tests fail here for
the expected reason: the pinned core (`5039f7e`) contains no `wreal` at all —
no `KwWreal` token, no `NetType::Wreal`, no `RealSum` resolution — so
`wreal x;` degrades to a default, integer-valued net. That accounts for every
assertion in the failing job: values rounding, negatives wrapping unsigned,
drivers not summing. Detail in the comment below.

### Verified locally against the core branch

CI runs two configurations, so both were checked, via
`scripts/use-local-core.sh`:

```
$ cargo test --release --test misc sv_compliance_runner
test sv_compliance_runner::test_sv_50_wreal_nets ... ok
test sv_compliance_runner::test_sv_neg13_wreal_packed_range ... ok
test result: ok. 69 passed; 0 failed; 1 ignored
```

```
$ cargo test --release --features jit --test misc wreal
test sv_compliance_runner::test_sv_50_wreal_nets ... ok
test sv_compliance_runner::test_sv_neg13_wreal_packed_range ... ok
test result: ok. 2 passed; 0 failed
```

The JIT needs no separate `wreal` work: it shares the same value
representation, so core's `RealSum` resolution carries through both paths.
Worth stating explicitly, since a reasonable reading of the failures is that
the JIT keeps its own representation and would need its own handling.

### Note on the test number

Originally written as `49_wreal_nets.sv`. `49_` was taken by
`49_specparam_hierarchical.sv` in the meantime, so this is renumbered to
`50_` on rebase.

### DIFF
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -380,3 +380,7 @@ fn test_sv_neg10_duplicate_enum_literal() {
 fn test_sv_neg11_missing_named_port() {
     run_negative_compliance_test("neg11_missing_named_port.sv");
 }
+#[test]
+fn test_sv_neg13_wreal_packed_range() {
+    run_negative_compliance_test("neg13_wreal_packed_range.sv");
+}
--- a/tests/misc/sv_compliance_runner.rs
+++ b/tests/misc/sv_compliance_runner.rs
@@ -312,6 +312,10 @@ fn test_sv_49_specparam_hierarchical() {
     run_positive_compliance_test("tests_advanced", "49_specparam_hierarchical.sv");
 }
 #[test]
+fn test_sv_50_wreal_nets() {
+    run_positive_compliance_test("tests_advanced", "50_wreal_nets.sv");
+}
+#[test]
 fn test_sv_41_genfor_localparam_idx() {
     run_positive_compliance_test("tests_advanced", "41_genfor_localparam_idx.sv");
 }
--- a/tests/sv_compliance/manifest.csv
+++ b/tests/sv_compliance/manifest.csv
@@ -49,3 +49,4 @@ file,topic,description
 46_nettype_hierarchy.sv,user-defined nettypes across hierarchy,nettype nets driven from several module instances through ports; Millman resolution over a nested dut catches per-port double resolution
 47_nettype_coercion.sv,nettype coercion of implicit nets,implicit net used as a port actual takes the port user-defined nettype (§6.6.8/§23.3.3)
 49_specparam_hierarchical.sv,specparam hierarchical access,specparam is a module-scoped elaboration constant reachable by hierarchical name (§6.20.5/§23.3.3) including through generate scopes and specify-block uses
+50_wreal_nets.sv,Verilog-AMS wreal nets,net whose value is a real with multiple drivers resolved by summing (KCL current summing); scalar and port forms
--- a/tests/sv_compliance/tests_advanced/50_wreal_nets.sv
+++ b/tests/sv_compliance/tests_advanced/50_wreal_nets.sv
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: MIT
+//
+// 50_wreal_nets.sv — Verilog-AMS `wreal`: a net whose value is a real, with
+// multiple drivers resolved by SUMMING them.
+//
+// `wreal` is a NET TYPE that names no data type of its own, so a signal
+// declared with it has to be given real storage from the net type alone.
+// Declared as a plain net it lands as a bit vector and every value is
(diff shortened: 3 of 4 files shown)


vrajeshprakhya:
The two failures here are a stale dependency pin, not a defect in the
simulator's value handling. Worth saying explicitly, because the symptom set
reads like an integer-coercion bug and invites a fix in the wrong place.

### Why CI fails

`Cargo.toml` pins `xezim-core` by exact rev:

```
xezim-core = { git = "https://github.com/aionhw/xezim-core.git", rev = "5039f7e..." }
```

`.cargo/config.toml` — the local-checkout patch that `scripts/use-local-core.sh`
writes — is gitignored, so CI never sees it and resolves core from git at that
rev. And `5039f7e` contains no `wreal` support at all:

```
$ git grep -i wreal 5039f7e -- src xezim-parser
… KwReal, KwRealtime, and nothing else
```

No `KwWreal` token, no `NetType::Wreal`, no `RealSum` resolution. Those are
added by aionhw/xezim-core#36, which is open and unmerged.

So CI compiles `wreal x;` with a parser that does not know the keyword. It
degrades to a default net, a default net is integer-valued, and every reported
symptom follows from that one fact:

- `2.5` rounding to `3.0` — integer storage
- negatives wrapping to a huge unsigned value — unsigned integer storage
- drivers not summing as reals — the `RealSum` resolution does not exist yet
- fractions lost across a module port — the port carries the same degraded net

Seven assertions, one missing feature.

### The same tests pass when core has it

Built against the core branch via `scripts/use-local-core.sh`:

```
test sv_compliance_runner::test_sv_50_wreal_nets ... ok
test sv_compliance_runner::test_sv_neg13_wreal_packed_range ... ok
test result: ok. 69 passed; 0 failed; 1 ignored
```

Identical simulator code and identical tests — only the core dependency
differs. That rules out a value-layer bug on this side.

### Please don't "fix" the value layer

A suggested remedy that came out of triage was to rework net/value storage to
`f64` and add real multi-driver summation here. That work already exists, in
`elaborate.rs` on the core branch:

```rust
NetType::Wreal => Some(Self::RealSum),
```

Doing it again in the simulator would duplicate the feature and hide the fact
that the dependency is stale.

### What actually unblocks this

1. Merge aionhw/xezim-core#36 (green, `MERGEABLE`, `CLEAN`)
2. Bump the pinned rev in `Cargo.toml` to the resulting core `main`
3. This PR goes green

No change is needed to this PR itself. Flagged in the description when it was
opened; posting the evidence here so it is on the failing PR rather than only
in the paired one.

aionhw (approved):
Reviewed together with xezim-core#36 and verified end-to-end against current `main` on both sides: core branch applied to `5039f7e` (clean), this branch merged with xezim `main` (clean, including the `49_` → `50_` renumber against `49_specparam_hierarchical`), built against the patched core — **`test_sv_50_wreal_nets` and `test_sv_neg13_wreal_packed_range` both pass, full merged-tree suite 2340/0, c906 gate bit-exact**. Registration is complete this time in all three places (runner, manifest, file) — `tests_advanced/` is at 32 files / 32 registrations after this.

The tests themselves are the strong part of the pair:

- **Case F (single driver passes through, not doubled) is the assertion I would have asked for** — it pins the fold's identity element, which every multi-driver case is structurally blind to. Same instinct as the three-driver sum being checked against the arithmetic total so last-write-wins cannot pass by luck.
- The all-fractional values mean the most likely failure mode (silent degradation to an integer net) trips every assertion rather than passing accidentally — and that is exactly what the currently-pinned core does with them, which the PR's own red-CI analysis documents accurately.
- `neg13` rides the line-anchored bare-`error:` matcher that #136 fixed, and works.

One observation for the record rather than the author: with this and core#36 in, a `wreal` node and a `§6.6.7 nettype real ... with sum` node agree on what several drivers mean, and the resolver-compilation from #137 applies to the folded `+` chains unchanged — the AMS pieces are converging on one arithmetic path rather than three.

**Approve**, contingent only on ordering: xezim-core#36 first, pin bump, then this — CI here cannot go green before the pinned core knows the keyword, exactly as the PR body says.

PR #141, opened 2026-08-27
Fix SystemVerilog semantics: statics, $typename, %p signedness, param bindings

Thomasb81:
Fixes a set of SystemVerilog semantics bugs, each with a self-test that produces byte-for-byte-identical output to the reference simulator. Also rebased onto current main (drops a redundant xezim-core pin bump that main already carries) and removes remaining uvm-tests test-path references from commit messages and committed comments.

### Changes

- Resolve built-in process class state enum constants inside subroutine bodies
- Resolve package data members (const / variable) inside subroutine bodies
- Give parameterized-base statics per-specialization storage for concrete subclasses
- Make per-specialization statics a generic receiver-derived mechanism
- Resolve class type-parameter formals before casting in socket construction
- Wake level-sensitive wait at the satisfying delta before an inactive #0 clobber
- Restore enclosing assoc-array registration after a nested same-named ref formal clears it
- Size a class-field operand to its declared width in a comparison's self width
- Classify a block-local QUEUE declaration as a queue var
- Resolve an enclosing value param in a nested generic's type-arg binding
- Parameterized-event callbacks, static-task spec, and single-eval method width
- `$typename` of a type parameter reports the concrete bound type
- `$typename` preserves packed-vector ranges and base types
- Fix static collection reuse of parameterized classes
- Fix explicit-spec static-collection reads for parameterized classes
- Fix `%p` on subroutine locals and string truncation through type params
- Fix signedness of type-param locals/formals in `%d`/`%p`

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -84666,11 +84666,20 @@ impl Simulator {
                     // is active, each specialization gets its own static cell.
                     // This applies to statics declared in the spec's base class
                     // AND inherited from parameterized ancestors (per IEEE 1800-2023 §8.25).
-                    let active_spec = spec_override.or(self.current_spec.as_ref());
+                    // Fall back to synthesizing a spec from a CONCRETE receiver
+                    // (e.g. `ext1::ID()` where `ext1 extends base#(ext1)`) so any
+                    // static reachable through such a subclass is keyed per
+                    // specialization too — this is the generic home of the
+                    // concrete-receiver path, so it applies uniformly to static
+                    // calls, instance-inherited statics, and property reads.
+                    let active_spec = spec_override
+                        .or(self.current_spec.as_ref())
+                        .cloned()
+                        .or_else(|| self.derive_static_spec(start_class, &cname));
                     let key = match active_spec {
                         Some((base, sig))
-                            if (base == &cname
-                                || self.class_extends(base.as_str(), &cname))
+                            if (base == cname
+                                || self.class_extends(&base, &cname))
                                 && self.class_is_parameterized(&cname) =>
                         {
                             // For inherited statics (cname != base), derive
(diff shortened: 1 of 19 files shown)


aionhw (approved):
Reviewed and verified beyond CI (which is green on both checks):

- **Merged with current `main` locally** (no conflicts, including with this week's NBA last-write-wins and compare-extension work) and ran the full suite on the merged tree: **2338 passed / 0 failed**. c906 CoreMark gate bit-exact on the PR build.
- **UVM regression sweep** (apb / axi4 / i3c AVIPs), PR build vs current main: apb and axi4 byte-identical; i3c differs only by a NEW `[warning] enum member 'ADDRESS' redefined with a different value` — an accurate §6.19 diagnostic on a real name collision in that bench, not a behavior change.
- **The scheduling commit got the closest look** (`Wake level-sensitive wait at the satisfying delta before an inactive #0 clobber`) since it touches region ordering. The structure is right: the existing end-of-tick condition-waiter fixpoint moves verbatim into `drain_condition_waiters()`, with one new call site before the inactive promotion. My one concern was an NBA queued by a freshly-resumed waiter committing before the `#0` region (a §4.9.4 ordering question); I built that exact scenario — `x=1; #0 seen=q` racing `wait(x==1); q<=5` — and the reference simulator, current main, and this PR all print `seen=0 q_after=0`. Concern retired empirically.
- **`class_time_field_neg_one` deserves a note**: it lands squarely next to this week's `is_equal` extension-rule fix (`c0f9619`/`c15c8c3`) and is complementary, not overlapping — that fixed how operands EXTEND, this fixes the literal never being SIZED to a class field's width because `lrm_self_width` cannot see per-object storage. Both suites' pins pass together on the merged tree, which is exactly the cross-check I wanted.

Seventeen semantics fixes each carrying a reference-verified self-test is the right shape for a batch like this, and the commit-message hygiene requested earlier (test-path references) is done.

**Approve** — merging.

Issue #38, opened 2026-07-19
[Suggestion] Thoughts on CI automation, compliance roadmaps, cross-platform tooling, and next-gen AI-driven debugging for Xezim


opensource-elearning:
Hi @aionhw  / @Thomasb81 / @oscargus / @rpjayaraman / @Ganesh-TNL

First of all, thank you for the incredible work on Xezim. Building an independent SystemVerilog simulator in Rust and reaching this level of functionality—especially with UVM support—is a remarkable achievement. The project's core vision of exploring AI-assisted EDA development to lower the barrier to entry is highly inspiring, and it’s exciting to see what a lean approach can accomplish.

I wanted to open this issue to share a few friendly suggestions and long-term feedback from an outside perspective. I realize these ideas touch on a wide range of features that might look overwhelming at first glance, but given how rapidly Xezim has progressed to achieve UVM parity with Verilator, it feels completely achievable for this architecture. 

Please feel free to take, leave, or modify any of these ideas—they are entirely up to your discretion and what fits best with your personal vision for the project!

---

### 1. Strengthening Continuous Integration (Locking in Progress)

As the project scales, it might be incredibly helpful to protect all the amazing progress you’ve already made by establishing a simple GitHub Actions workflow on pushes and pull requests. 

This pipeline could handle:
* **Basic Linting:** Automated `cargo fmt --check` and `cargo clippy` runs.
* **Regression Protection:** Running all the existing unit tests, integration tests, and UVM test suites that are currently passing.

Having all currently passing tests run automatically in CI ensures that future additions or experiments don't accidentally introduce regressions to existing features, giving you a completely stable foundation to build on.

### 2. Leveraging `sv-tests` as a Living Roadmap

Instead of spending precious time maintaining a manual TODO list, integrating the [CHIPS Alliance SV-Tests compliance suite](https://github.com/chipsalliance/sv-tests) into the workflow could be a massive time-saver. 

Running Xezim against it would give you a clear, automated matrix showing exactly where things stand:
* Language features passed vs. failed.
* Parser gaps and unsupported syntax constructs.
* Semantic and elaboration discrepancies.

This compliance matrix naturally becomes an effortless living roadmap, making it easy to see what language features are missing and allowing you to triage or prioritize them whenever you feel like tackling them.

### 3. Differential Testing for Behavioral Correctness

[...]

PR #148, opened 2026-08-31
Parameterized-class static collections, method binding, and reflection follow-ups

Thomasb81:
## Parameterized-class static collections, method binding, and reflection follow-ups

Follow-up to #139: continued SystemVerilog semantics fixes in the simulation runtime, all validated byte-for-byte against a reference simulator where stated. The commits form one interdependent chain, so they land together.

Highlights:

- **Per-specialization static collections**: same-named static collections in sibling parameterized classes get independent storage (only sibling classes split the storage); qualified static-collection access through a typedef-specialization alias now resolves the same per-spec cell from module scope as in-method access; class-qualified element reads and builtin calls (`EMPTY`/`FULL`/`NUM`) work from non-instance contexts; static per-specialization collections resolve consistently.
- **Whole-collection semantics**: a `ref` formal shadows a same-named class property for whole-collection copies; whole-collection LITERAL returns (`q = f()`) snapshot the collection; an element of a class-member string collection is treated as a string.
- **Method binding and registry**: a bare parameterless instance method referenced by name invokes (§13.4.1); a non-virtual method no longer binds to a static member's declaring class; `C::type_id::create` routes through the override-aware registry path; parameterized-class registry dispatch works through typedef aliases.
- **Enum reflection**: `.name()` resolves a method-local enum type from the active call frame, not the flat per-name maps; `$cast` to a class-member/method-local enum range-checks.
- **Randomization**: an unseeded object's `randomize()` draws from the process stream.
- **Misc**: frame-scoped string-flag registry so a nested non-string local can't clobber an enclosing `string` declaration; §6.18 — assigning a non-class value to a class-handle variable is rejected at compile time; parameterized element specs bound and assoc-array declaration index props evaluated; call-bearing instance field initializers evaluated at construction time.

Adds 17 self-tests across `tests/types`, `tests/classes`, `tests/collections`, `tests/strings`.

Full self-test suite: 340 passed; the 2 failures are pre-existing environment-specific config-db integration tests, unrelated to this branch.

**Dependencies:** requires xezim-core PR **#38** (static assortment key metadata, package-nested base classes in constraint name checks, type-keyword specialization args on data declarations). The `Cargo.toml` core `rev` pin (currently `e15a759`) must be bumped to the merged core `main` revision once that lands.


Thomasb81:
Heads-up that this PR's head branch has been advanced from `96f2e9a` to `4c13736` to fold in the additional work that landed on `fix/followup-139` and the merged upstream `main`. The newly-added commit range is **`96f2e9a..4c13736`** (22 commits), most notably:

- **`7770500` fix: resume level-sensitive `wait` waiters in write-satisfaction order** — §9.7.4: a waiter satisfied by an *earlier* blocking write now resumes (when the writer yields) before one satisfied by a later write, instead of in FIFO parking order. Uses sensitivity-gated write-order tracking so full-scan cost is avoided. This also flips the `wait_satisfaction_order::earlier_satisfied_waiter_resumes_before_later_parked` regression guard to passing and fixes the driver/sequencer `item_done` handshake race.

- **`18e0d60` fix: preserve `static_local_syncs` in `ProcessContext` across process suspension** — §6.21/§13.3.1: a `static` subroutine local now latches correctly across a task suspension/resumption, so re-entrant scheduling loops no longer hang. Also converts `phase_jump_static_latch` into a pure-SV testbench (no UVM dependency).

- **`ea8b919` fix: keep cross-process same-named block locals distinct** — alpha-renames block locals declared by more than one process so sibling `initial` blocks no longer cross-read each other's values.

- **`f8744ad` Solver tight mode** plus the type/enum/foreach/hier test additions and the xezim-core bump that came in with solving/struct-port work.

- The `Merge remote-tracking branch 'origin/main'` (`4c13736`) reconciles this branch with the current upstream `main`, so the PR diff is based on a clean, up-to-date base.

aionhw (changes requested):
Built and tested the pair (this PR + xezim-core #38) trial-merged onto today's `main` (xezim 656860b / core ace27f3). Both merge clean; the merged tree builds; core suite clean; xezim default suite: all green except one of this PR's own new tests (details below). Review concerns were turned into repros and checked against the reference simulator.

**Requesting changes** for two reference-divergent defects and one measured perf regression; the rest of the PR looks right and is a clear net improvement (the static-collection split, factory routing, `.name()` frame scoping, `ref` formal shadowing, and the literal-return snapshot all check out).

### 1. Call-bearing field initializers run more than once per construction (`instantiate_class` fixed-point loop)

```systemverilog
class counter; static int next = 0;
  static function int take(); next++; return next; endfunction endclass
class item; int id = counter::take(); int id2 = counter::take(); endclass
// item a = new, b = new, c = new;
```
Reference: `a=1/2 b=3/4 c=5/6 next=6`. This PR: `a=5/6 b=11/12 c=17/18 next=18` — every initializer ran three times per construction (`0..=deferred_call_inits.len()` rounds, each re-evaluating every deferred initializer). §8.7: initializers run once, in declaration order, as if they were the first statements of `new()`. The fixed point exists only because `property_inits` is a `HashMap`; the fix is to keep declaration order in `ElaboratedClass` (a `Vec`) and evaluate each initializer exactly once, in order. (main today leaves all of them 0, so the singleton-getter case this targets is real — it just needs the ordered form.)

### 2. Core #38: a dimensioned vector type arg collapses to its keyword

`param_value_to_type_arg` renders `DataType::IntegerVector` as the bare keyword, dropping `dimensions`, so `P#(bit [7:0])` and `P#(bit)` become the same specialization:
```systemverilog
class P #(type T = int); function int width(); return $bits(T); endfunction endclass
P #(bit [7:0]) a; P #(bit) b; P #(int) c;
```
Reference `8 1 32`; with #38 `1 1 32` (main: `32 32 32`). Suggest returning `None` when `dimensions` is non-empty (keeps today's behaviour for that shape) or rendering the dims into the fragment.

### 3. +3.6% instructions on the axi4 AVIP

Interleaved `perf stat -e instructions`, `axi4_base_test`: main 43.477G / 43.478G, this PR 45.045G / 45.052G. `perf record` attributes it diffusely to three new hot-path costs:
- `string_signals_saves`: the whole `string_signals` `HashSet` is cloned on every function/method entry and restored on exit. Logging this frame's *removals* and re-inserting them on exit (the mirror of the existing `frame_string_signals` add-log) gives the same guarantee with no clone.
- `check_condition_waiters_for_write` runs inside `assign_value` whenever any `wait(cond)` waiter is parked (UVM's phase machinery keeps some parked), collecting lvalue names into a fresh `Vec<String>` per blocking write and scanning every waiter's read-set.
- `static_coll_name_collides` scans every class on each static-collection access (`spec_static_coll_key`, the receiver rewrites, the Index read path). The startup loop already computes exactly this set as `colliding_names` — cache it on the simulator and look it up.
Also visible: `fn_returns_collection` clones the return `DataType` per call, and the fixed-point loop in (1) clones `deferred_call_inits` per round.

### 4. `cross_process_same_named_frameless_local_stays_distinct_named_blocks` fails on the merged tree

On main + this PR the UNNAMED variant passes (`u1=20 u2=02`) but the named-block variant still collides (`r1=01 r2=01`, p1 reads p2's value). This is not a merge interaction: the same test fails at this PR's own head built against the core it pins (a7bf5ba) plus core #38 (`3 passed; 1 failed` in `process_block_local_shadowing`), and a standalone run gives `NAMED r1=01 r2=01 UNNAMED u1=20 u2=02`. So the unnamed-block alpha-rename works and the labeled-block one does not — the `frameless_locals` collector treats both the same, so the divergence is downstream (a labeled block's locals are registered/rewritten on a different path, presumably the statement-local scoping in core `d6e087f`). Worth fixing since the test documents it as reference-validated; otherwise drop the test until it does.

### Notes (no change requested)
- `cur_rng`: an object without its own stream now draws from the calling process's stream instead of the global default. LRM-consistent with thread attribution, but it changes every existing `+seed` sequence — worth a release note.
- The PR description says it depends on core #38 at pin `e15a759`; the branch (after its merge of main) actually needs core main at least at `8cb36ab` (`wide_enum_value_map`, `Value::set_scalar_code`) — the PR head does not build against core #38's own head. Please bump the pin to the merged core `main` in this PR.
- `is_type_two_state_resolved` (core #38): a base-less enum now defaults 2-state per §6.19 — correct, and it matches the reference (0, not x).
- The §6.18 non-class-to-handle compile check is deliberately conservative (top-module blocks, bare LHS). Fine as a first cut; sub-instance bodies are now covered by the design-wide identifier check that landed in core `ace27f3`, so a follow-up could route through that.

PR #149, opened 2026-09-03
fix: sequence body dispatch, #0 objection yield, and procedural NBA waiter scheduling

Thomasb81:
This PR resolves several interrelated issues affecting sequence dispatch, objection draining, and NBA synchronization:

1. **Inline blocking-method local scoping (`method_local_base`)**:
   - Inlined blocking task/method calls now track `method_local_base` like synchronous methods, and `local_class_type_of` / `local_typedef_type_of` are bounded to `method_local_base.last()`. This prevents an outer caller's local variables (e.g. `seq` in `start_phase_sequence` or `configure_phase`) from shadowing a sequence subclass's `seq` member property, ensuring `body()` dispatches to the subclass and `$cast` resolves to member properties rather than caller locals.
   - In `class_of_var`, properties of `this` are checked before falling back to the flat global `var_class_types` map.

2. **Yield level-sensitive condition waiters to pending `#0` continuations**:
   - Before promoting the Inactive (`#0`) region in `run_one_tick`, `drain_condition_waiters_pre_inactive` avoids committing NBA (IEEE 1800 §4.5) and yields after each round if continuations are pending in `inactive_queue`. This prevents multi-generation condition-waiter cascades (such as background objection drains) from starving `#0` continuations, allowing sequence completion (`#0` / join / cleanup) before phase-end termination guards fire.

3. **Procedural NBA waiter scheduling (`nba_region_waiters`)**:
   - When `@(var)` is waited upon for a procedural variable (as in `uvm_wait_for_nba_region`: `nba <= next_nba; @(nba)`), the process is parked in `nba_region_waiters` rather than scheduled immediately into the active event queue. The waiters resume in the Reactive region after all Active and Inactive (`#0`) deltas have settled and `apply_nba` has committed.
   - In `event_loop_singlethread`, pending NBAs and NBA waiters keep the event loop at the current timestamp rather than prematurely advancing time.

4. **Self-tests added**:
   - `nested_seq_method_dispatch`: verifies virtual dispatch of nested sequence `body()` methods under the default sequence runner.
   - `caller_local_shadows_this_cast`: validates that outer caller locals do not shadow instance properties in `$cast`.
   - `condition_waiter_yields_to_inactive`: validates that condition-waiter cascades yield to pending `#0` continuations per delta cycle.
   - `nba_wait_for_region_yield`: verifies that procedural NBA waiters yield until all inactive delta cycles at the timestamp complete.
   - `package_class_nested_class`: verifies accessibility and instantiation of nested classes declared inside package-scoped classes.
   - `sqr_zero_time_loop`: runs a real sequencer whose sequence `wait_for_relevant` forgot to wait (non-blocking, `is_relevant` stays 0), so the zero-time arbitration must spin until `SEQRELEVANTLOOP` fires.

5. **Consume same-tick-readied condition waiters**:
   - `drain_condition_waiters` (and the pre-inactive variant) bailed as soon as `cond_progress` had not advanced, even when a same-tick blocking write had moved a waiter into `ready_condition_waiters`. Because moving a waiter to ready does not itself fire its body, the readied waiter was stranded and its `join_any` parent stayed parked in `join_waiters`. This is exactly what the sequencer's `m_select_sequence` zero-time arbitration does (a fork child re-parks on `wait(m_is_relevant_completed > 0)` while a sibling writes the flag), so the arbitration loop ran only once instead of spinning to the `SEQRELEVANTLOOP` fatal. The drain only treats "no progress" as exhausted when `ready_condition_waiters` is also empty, so readied waiters are always consumed before the next tick. Verified byte-for-byte against reference simulators: 12 `wait_for_relevant` calls, `SEQRELEVANTLOOP` fatal at t=1, caught -> PASS.

Verified byte-for-byte against reference simulators.

**Dependencies:** requires xezim-core PR [**#39**](https://github.com/aionhw/xezim-core/pull/39) (register nested classes for package- and module-scoped class declarations). The `Cargo.toml` core `rev` pin (currently `0e27469`) must be bumped to the merged core `main` revision once that lands.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -4590,6 +4590,11 @@ pub struct Simulator {
     /// the event queue only at the END of the current tick, after
     /// `apply_nba` has run (see `promote_inactive_to_active`).
     inactive_queue: Vec<(usize, ProcCont)>,
+    /// Processes waiting for the NBA region to commit (e.g.
+    /// `uvm_wait_for_nba_region`: `nba <= next_nba; @(nba)` where `nba` is a
+    /// procedural variable). Resumed in `run_one_tick` after all active +
+    /// inactive (`#0`) deltas have settled and `apply_nba` has committed.
+    nba_region_waiters: Vec<(usize, ProcCont)>,
     /// Bumped whenever a `wait(expr)` condition evaluates true (a parked waiter
     /// proceeds). The condition-waiter fixpoint uses it to detect "no progress
     /// this round" and stop (genuine stall).
@@ -8217,6 +8222,7 @@ impl Simulator {
             cond_waiter_reads: HashMap::default(),
             cond_waiter_conditions: HashMap::default(),
             inactive_queue: Vec::new(),
+            nba_region_waiters: Vec::new(),
             cond_progress: 0,
             real_time: 0.0,
             mailbox_get_waiters: HashMap::default(),
@@ -34805,6 +34811,7 @@ impl Simulator {
         // region — so they observe post-edge state and this cycle's samples.
         self.drain_deferred_clocking_conts();
         self.drain_reactive_region();
+        self.drain_nba_region_waiters();
         self.fire_vpi_synch_cbs();
         self.check_monitor();
         self.drain_pending_strobes();
@@ -35287,10 +35294,19 @@ impl Simulator {
                 None
             };
             let next_delayed = self.next_delayed_time();
+            let next_nba_time = if !self.nba_fast.is_empty()
+                || !self.nba_queue.is_empty()
+                || !self.nba_region_waiters.is_empty()
+            {
+                Some(self.time)
+            } else {
+                None
+            };
             let next_time = [
                 next_eq_time,
                 next_clk_time,
                 next_delayed,
+                next_nba_time,
                 self.next_vpi_cb_time(),
             ]
                 .into_iter()
@@ -39519,9 +39535,10 @@ impl Simulator {
                                 // waited on (uvm_wait_for_nba_region:
                                 // `nba <= next_nba; @(nba)`). Its event_waiter
                                 // would resolve to no signal_id and park forever;
-                                // treat it as a one-delta yield (its purpose is to
-                                // yield across the NBA region).
-                                self.event_queue.schedule(self.time, pid, cont);
+                                // park in nba_region_waiters so it resumes only
+                                // after the active + inactive (#0) deltas have
+                                // settled and the NBA region has committed.
+                                self.nba_region_waiters.push((pid, cont));
                             }
                             return;
                         }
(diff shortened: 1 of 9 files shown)


aionhw (changes requested):
Built and tested the pair (this PR + xezim-core #39) trial-merged onto today's `main` (xezim 3aaf373 / core 8b2a904). Both merge clean, the merged tree builds, and the core suite, the full default suite, the JIT suite, `classes` (317) and `scheduling` (336) are all green. All six AVIPs match their baselines on the PR build (axi4 UVM_ERROR 0 @3100, apb 4 @1010, spi 1 @100, i3c 2 @3010, ahb 4 @40, uart to max-time). The dependency on core #39 is real: the `package_class_nested_class` design prints `FAIL` on core `main` and `PASS val=123` with #39, so core merges first and this PR carries the pin bump.

The three fixes read correctly, and I checked the part with the most room for an off-by-one: both new `method_local_base` pushes happen after `bind_task_frame` has pushed the callee's locals frame (simulator.rs:99320), so `local_stack.len()-1` is the index *of* that frame — the same convention as the synchronous site — and the push/pop pairing balances through `pushed_method_this`. The reference simulator also agrees with the new tests' expectations (`CAST_SUCCESS val=42`, `STAGE_COMPLETED stage=1`, `PASS val=123`, `ZERO_DELAY_DONE` before `WOKE_AFTER_NBA`), and the `uvm_wait_for_nba_region` idiom three times in a row completes at t=1 just as the reference does.

**Requesting one change**, plus three notes.

### `method_local_base` is simulator-global but is not part of `ProcessContext`

`ProcessContext` carries `this_stack`, `local_stack`, `local_type_stack`, `class_context_stack`, `task_cleanup` and `static_local_syncs` — but not `method_local_base`. Until now the base was pushed only on the synchronous `exec_method_in_class_hierarchy` path; this PR pushes it on the inlined **blocking** method path, which is exactly where the process is expected to suspend. When a process parks inside such a method, its `local_stack` is `mem::take`n into its `ProcessContext` while its base stays on the simulator, so another process's `local_class_type_of` / `local_typedef_type_of` / `in_any_frame` end up bounded by a foreign index.

```systemverilog
class Wrong; int tag; function new(); tag = 111; endfunction endclass
class Right; int tag; function new(); tag = 222; endfunction endclass
class Parker;
  task blocking_wait();
    Right obj;              // method-local shadows the module-scope Wrong obj
    #5;                     // parks: base pushed, context swapped out
    obj = new();
    $display("PARKER tag=%0d (expect 222)", obj.tag);
  endtask
endclass
class Deep; task go(); #20; endtask endclass
module tb;
  Wrong obj;                // module-scope net of a different class
  Parker p = new; Deep d = new;
  task automatic lvl2(); d.go(); endtask   // entered with 2 caller frames below
  task automatic lvl1(); lvl2(); endtask
  initial p.blocking_wait();
  initial lvl1();
  initial #40 $finish;
endmodule
```

Reference simulator: `PARKER tag=222`. `main`: `222`. This PR: **`111`**. `Deep::go`'s base (2) outlives its owner while the parker resumes with a one-frame stack, so `local_stack.get(2..)` is empty, `in_any_frame` is false, `obj` resolves to the module-scope net and the wrong class is constructed. Flattening the design so both processes sit at the same frame depth prints `222` on this build, which pins the mechanism to the depth mismatch rather than anything else.

Fix: add `method_local_base` to `ProcessContext` and to snapshot/take/restore, the same shape `static_local_syncs` got in #148 — or key it per pid. Worth noting neither the full suite nor the six AVIPs catch this, so a regression test alongside the fix would be welcome.

### Notes

1. **+0.42% instructions on the axi4 AVIP** (interleaved `perf stat -e instructions`: main 44.346G / 44.338G, this PR 44.530G / 44.524G — consistent across both rounds). `perf record` puts `class_prop_type_named` at 0.24% self on the PR build and below 0.05% on main, with `String::clone` at 0.35%: that is the new `this`-property probe in `class_of_var` and its `class_name.clone()`. Borrowing the class name instead of cloning it, and skipping the probe when the flat map already yields a known class, should recover most of it.
2. `drain_condition_waiters_pre_inactive` duplicates about fifty lines of `drain_condition_waiters`. A `pre_inactive: bool` parameter would keep the two from drifting — this PR already had to land the same "consume readied waiters" fix in both copies, which is the drift starting.
3. `report_parked_waiters`, the `--max-time` hang report, does not include the new `nba_region_waiters` lane, so a hang with a process parked there prints nothing about it.

Issue #40, opened 2026-07-20
event Data Type and Related Features


Ganesh-TNL:
Sample test case on EDA Playground for the SV event data type (Section 6.17) and its functionality (Section 15.5) of IEEE Std. 1800-2017

https://www.edaplayground.com/x/ZUDm

<details>

<summary>SystemVerilog Code for event Test Case</summary>

```systemverilog
`timescale 1ns/1ps

// ============================================================================
// 1. SELF-CHECKING PREAMBLE MACROS
// ============================================================================
`ifndef SVTEST_DEFS_SVH
`define SVTEST_DEFS_SVH

`define SVTEST_INIT \
  int failures = 0;

`define SVTEST_CHECK(expr, msg) \
  if (!(expr)) begin \
    failures++; \
    $display("FAIL: %s at time %0t ps", msg, $time); \
  end

`define SVTEST_PASSFAIL \
  if (failures == 0) begin \
    $display("TEST_PASS"); \
  end else begin \
    $display("TEST_FAIL count=%0d", failures); \
    $fatal(1); \
  end

`endif

module sv_event_tests_tb ;

  // Global monitoring registers
  `SVTEST_INIT
  // --- Infrastructure for Clocking Block Event Testing (Sec 14.13) ---
  reg clk_cb;
  initial clk_cb = 1'b0;
  always #5 clk_cb = ~clk_cb; // Generate a standard 10ns clock

  // Clocking block using an event control expression (Sec 14.13)
  default clocking cb @(posedge clk_cb);
    default input #1ps output #1ps;
  endclocking
  
  // Named event declarations (Sec 6.17)
  event ev_sync_a;
  event ev_sync_b;
  event ev_sync_c;

  // Global monitoring flags for task/function testing
  bit task_call_1_triggered = 0;
  bit task_call_2_triggered = 0;

  // ========================================================================
  // AUTOMATIC TASK WITH LOCAL EVENT (Sec 15.5 / Sec 13.4.2)
  // Each invocation of an automatic task creates a unique, isolated local
  // instance of the event variable and its underlying synchronization object.
  // ========================================================================
  task automatic run_automatic_event_task(int instance_id, int delay_ps, output bit out_triggered);
    event local_auto_ev; // Created dynamically on stack per call
    out_triggered = 0;

    fork
      begin
        #(delay_ps);
        -> local_auto_ev;
      end
      begin
        #(delay_ps);
        if (local_auto_ev.triggered) begin
          out_triggered = 1;
        end
[...]

PR #150, opened 2026-09-03
fix: carry method_local_base through ProcessContext context swaps (and address PR #149 review notes)

Thomasb81:
Addresses the review of xezim PR [**#149**](https://github.com/aionhw/xezim/pull/149): the requested change plus the three non-blocking notes.

### 1. Requested change — `method_local_base` was not part of `ProcessContext` (fixed)

`method_local_base` records the index into `local_stack`/`local_type_stack` that bounds `local_class_type_of` / `local_typedef_type_of` / `in_any_frame` to the current inlined method's own frames. It lived on the simulator globally while every other per-call frame walked with `local_stack` inside `ProcessContext`. When a process parked (`#5`) inside an inlined blocking method, its `local_stack` was `mem::take`n into its `ProcessContext` but its base stayed behind, so whichever process ran next left a foreign base index behind. On resume the parker bounded its restored one-frame stack with that stale index, `in_any_frame` turned false, and a method-local `Right obj` resolved to a same-named module-scope net of a *different* class — constructing the wrong type.

```systemverilog
class Wrong; int tag; function new(); tag = 111; endfunction endclass
class Right; int tag; function new(); tag = 222; endfunction endclass
class Parker;
  task blocking_wait();
    Right obj; // method-local shadows the module-scope Wrong obj
    #5;        // parks: base pushed, context swapped away
    obj = new();
    $display("PARKER tag=%0d (expect 222)", obj.tag);
  endtask
endclass
class Deep; task go(); #20; endtask endclass
module tb;
  Wrong obj; Parker p = new; Deep d = new;
  task automatic lvl2(); d.go(); endtask
  task automatic lvl1(); lvl2(); endtask
  initial p.blocking_wait();
  initial lvl1();
  initial #40 $finish;
endmodule
```

Reference simulator: `PARKER tag=222`. Before: **`111`**. After: `222`. **Fix:** move `method_local_base` into `ProcessContext` (snapshot/take/restore), mirroring `static_local_syncs` from #148. Adds regression self-test `blocking_method_keyed_method_local_base`.

### 2. Note 2 — dedup `drain_condition_waiters` (no more drifting copies)

`drain_condition_waiters_pre_inactive` duplicated ~50 lines of `drain_condition_waiters`, and the "consume readied waiters" fix already had to be landed in both copies (that was the drift starting). Collapsed into one `drain_condition_waiters(pre_inactive: bool)` that selects the two genuinely-different behaviors: the pre-INACTIVE(`#0`) drain skips the NBA flush (§4.5 — the NBA region stays strictly after the inactive region) and yields to a pending `#0` continuation; the end-of-active drain flushes pending NBA and runs the full same-time fixpoint.

### 3. Note 3 — `report_parked_waiters` now reports the `nba_region_waiters` lane

The `--max-time` hang report didn't include the `nba_region_waiters` lane, so a hang with a process parked on `@(procedural_var)` for the NBA region (e.g. `uvm_wait_for_nba_region`'s `nba <= next_nba; @(nba)`) printed nothing. Now counted in the early-return guard and reported per parked pid.

### 4. Note 1 — drop the `this`-property probe's per-lookup `String` clone

`class_of_var`'s `this`-property probe cloned the instance's class name on every lookup (`i.class_name.clone()`), ~0.35% `String::clone` on the axi4 AVIP. Borrow `&inst.class_name` instead, so `class_prop_type_named` allocates nothing on the hot class-method path.

Depends on xezim-core PR for the `register_nested_classes` single-elaboration change (so the core rev pin in `Cargo.toml` can be bumped together); a local-core-patched build passes `package_class_nested_class` and the full suite.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -2583,6 +2583,14 @@ struct ProcessContext {
     // `push_queue_frame`/`pop_and_restore_queue_frame`.
     local_dyn: Vec<Vec<(String, String)>>,
     static_local_syncs: Vec<(String, Vec<(String, String)>)>,
+    // Base index into `local_stack`/`local_type_stack` of THIS method's own
+    // locals frame, so that after a process parks inside an inlined blocking
+    // method and its `local_stack` is swapped into its `ProcessContext`, the
+    // bounds for `local_class_type_of`/`local_typedef_type_of`/`in_any_frame`
+    // follow the frame rather than leaking a foreign index from whichever
+    // process runs next. Kept in `ProcessContext` (like `local_stack` itself)
+    // precisely because the base is only meaningful against that stack.
+    method_local_base: Vec<usize>,
 }
 
 #[derive(Debug, Clone)]
@@ -36767,6 +36775,7 @@ impl Simulator {
             task_cleanup: self.task_cleanup.clone(),
             local_dyn: self.local_dyn.clone(),
             static_local_syncs: self.static_local_syncs.clone(),
+            method_local_base: self.method_local_base.clone(),
         }
     }
 
@@ -36793,6 +36802,7 @@ impl Simulator {
             task_cleanup: std::mem::take(&mut self.task_cleanup),
             local_dyn: std::mem::take(&mut self.local_dyn),
             static_local_syncs: std::mem::take(&mut self.static_local_syncs),
+            method_local_base: std::mem::take(&mut self.method_local_base),
         }
     }
 
@@ -36813,6 +36823,7 @@ impl Simulator {
         self.task_cleanup = ctx.task_cleanup;
         self.local_dyn = ctx.local_dyn;
         self.static_local_syncs = ctx.static_local_syncs;
+        self.method_local_base = ctx.method_local_base;
     }
 
     fn inherit_current_process_context(&mut self, pid: usize) {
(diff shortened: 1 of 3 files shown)


aionhw (commented):
Thanks for following the #149 review through, and for the reproducer.

**1. `method_local_base` in `ProcessContext`** — this exact fix landed on `main` right after the #149 review (the base is snapshotted, taken and restored with the rest of the context, and per-process). Your `PARKER` reproducer prints `tag=222` on current `main`, and your regression test passes unchanged. So this part of the PR is superseded rather than merged.

**2–4. Drain de-duplication, hang-report NBA-region lane, `this`-probe borrow** — all three are still applicable to `main` and are good changes. Since the branch is based on a `main` from 2026-09-03 and now conflicts in `simulator.rs`, I cherry-picked your second commit onto `main` with its authorship intact instead of asking for a rebase: `311e82d`. It applied cleanly, and the classes, scheduling, misc and hierarchy groups plus the full default and JIT suites pass on top of today's `main`. The parameterised `drain_condition_waiters(pre_inactive)` matches what the two copies had drifted into (NBA flush only on the end-of-active drain, `#0` yield only on the pre-inactive one), so behaviour is unchanged.

One sibling your fix pointed at: a subroutine-local `virtual ifc lv; lv = vi;` was recorded in the simulator-global alias map by bare name, so two class tasks interleaved on delays saw each other's binding (`a1`'s `lv.id` read `a2`'s interface after a park). That is the same "per-process state living on the simulator" class as `method_local_base`; it is fixed on `main` alongside this, with a test, and you are credited in the README contributors list for this round.

On the core side, PR aionhw/xezim-core#40 is reviewed separately. Closing this PR since everything in it is now on `main`.

aionhw:
Part 1 superseded on main; parts 2–4 landed as 311e82d with your authorship. Closing.

PR #151, opened 2026-09-03
Report concurrent cover properties as covers, not as failing asserts

vrajeshprakhya:
A `cover property` is recorded as an assertion, and every clock where it
does not hit is tallied as a **failure**. Two covers and nothing else:

```
[COV] assertions: 2 sites (assert=2, assume=0, cover=0) — 3 passes, 3 fails
```

Both halves are wrong. The second is the one that bites: an uncovered
property is not a failing property, so a coverage report built from
`xezim_cov.json` reads a healthy run as a regression.

## Cause

`SvaClockedSite` carries no kind, so every `AssertionStat` created while
evaluating a clocked property is built with `kind: 0 /* assert */`
hardcoded — six such sites across `tick_sva_sites` and `tick_sva_body` —
even though the immediate path already maps `a.kind` onto the same `u8`
tag correctly a few hundred lines away. The concurrent-cover probe path
notes the gap in a comment: *"reuse the immediate-cover tag for
concurrent-cover probes (no dedicated tag yet)"*.

## Fix

Carry the source kind on the site, set from `a.kind` at construction with
the mapping the immediate path already uses, and tally under it. Then
guard the fail increments: for kind 2 a non-hit increments nothing, since
a cover has hits rather than verdicts (LRM §16.12).

## Result

```
cover only              assert=0 cover=2  —  3 passes,  0 fails
3 asserts + 2 covers    assert=3 cover=2  — 13 passes,  5 fails
always-false assertion  assert=1 cover=0  —  0 passes,  4 fails
```

The last line is the control: real assertion failures still count and
still fire their `else` actions.

## Scope

Behaviour is unchanged — only the tallies move. Verified across a battery
of immediate asserts, implication failures (`a |=> b`), `disable iff`,
`$rose`, multi-cycle sequences, and `cover property` action blocks: every
one behaves exactly as before, and `xezim_cov.json` entries now carry
`"kind": "cover"` with `fail: 0`.

Built and tested on current `main`.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -2473,6 +2473,12 @@ struct SvaClockedSite {
     /// and swapped in while this site's predicate is evaluated, so the
     /// property samples pre-edge values (see `eval_sva_sampled`).
     sampled_ids: Vec<usize>,
+    /// LRM §16.5 assertion kind, same u8 tagging the immediate path uses:
+    /// 0 assert, 1 assume, 2 cover. Carried here because the tally is
+    /// written long after the statement is gone, and a `cover property`
+    /// recorded as an assert both mislabels the site and counts a cycle
+    /// that simply did not hit as a failure.
+    kind: u8,
 }
 
 #[derive(Debug, Clone)]
@@ -63918,6 +63924,11 @@ impl Simulator {
                                 pass_action: a.action.as_deref().cloned(),
                                 fail_action: a.else_action.as_deref().cloned(),
                                 sampled_ids,
+                                kind: match a.kind {
+                                    AssertionKind::Assert => 0,
+                                    AssertionKind::Assume => 1,
+                                    AssertionKind::Cover => 2,
+                                },
                             });
                         }
                         return;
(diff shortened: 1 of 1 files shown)


aionhw (commented):
Thanks for the careful diagnosis; the report is exactly right about the cause (no kind on the clocked site, so every tally was filed as `assert`, and a non-hit cover was counted as a failure).

This landed on `main` independently before the PR could be rebased: `SvaClockedSite` now carries the source kind and the main tally paths use it (part of the assertion round on 2026-09-04, in the `c940b05` lineage). On current `main` your two examples give:

```
cover only              [COV] assertions: 2 sites (assert=0, assume=0, cover=2) — 1 passes, 0 fails
1 assert + 1 cover      [COV] assertions: 2 sites (assert=1, assume=0, cover=1) — 2 passes, 3 fails
```

with the `else` action of the failing assert still firing and the cover's action firing on its hit.

Three sites your patch also covered were still hardcoded on `main` after that round: the `s_eventually` / `s_always` watcher tallies in `tick_sva_sites` and the vacuous-pass tally of `lhs |-> rhs` in `tick_sva_body`. I have folded those in on `main` following your diff (kind from the site, no fail increment for a cover), so nothing from this PR is lost.

Because the branch is based on a `main` from two days ago and now conflicts in `simulator.rs`, and the change is fully on `main`, I am closing this rather than asking for a rebase. Much appreciated.

aionhw:
Superseded by the fix on main (see review); closing.

PR #156, opened 2026-09-05
Two array-shape resolution fixes: unit-inner packed 2-D continuous assign, and handle-qualified array queries

chenbhil:
# Two array-shape fixes: unit-inner packed 2-D continuous assign, and handle-qualified array queries

Branch: `chenbhil/xezim:fix/packed-elem-assign-and-handle-array-query` -> `aionhw/xezim:main`

Two independent fixes, both inside `xezim`. No `xezim-core` change, no pin bump.
Both turned up running a FlooNoC router UVM smoke bench against Verilator
5.050. The first one blocked that bench.

## 1. `assign v[p][0] = x` on `logic [N-1:0][0:0] v` drives only index 0

`src/compiler/bytecode.rs`

```systemverilog
logic [4:0][0:0] v;          // packed 2-D, unit inner dimension
assign v[0][0] = 1'b1;  assign v[1][0] = 1'b0;  assign v[2][0] = 1'b1;
assign v[3][0] = 1'b1;  assign v[4][0] = 1'b0;
// before:           v = xxxx1
// Verilator 5.050:  v = 01101
```

A lone `assign v[3][0] = 1'b1;` gives `xxxx1`, so the write lands on bit 0, not
on element 3.

### Isolation

| Variant | Before |
|---|---|
| `assign v[p][0] = x` on `[N-1:0][0:0]` | wrong, only index 0 drives |
| same, explicit indices, no `generate` | wrong, so the genvar is not involved |
| `assign v[p] = x` (whole inner dim) | ok |
| `assign v[p] = 1'(x)` (cast form) | ok |
| `[N-1:0][1:0]`, 2-bit inner dim, `v[p][0]` | ok |
| plain 1-D `[N-1:0]`, `assign v[p] = x` | ok |
| procedural `v[3][0] = 1'b1` on `[N-1:0][0:0]` | ok |

Continuous assign only, unit inner dimension only.

### Cause

`compile_blocking_target`'s `ExprKind::Index` arm requires an `Ident` base, so
the nested `v[3][0]` skips it and falls through to
`flattened_outer_const_signal_id`, which returns the whole-vector signal id.
The caller then emits `BlockingAssignBitDyn(v, <inner index>)`, so every
`v[i][0]` writes bit 0 whatever `i` is.

That function already had the right guard:

```rust
// A multi-D PACKED base ... `foo[0][j]` must not degrade to a bit-select
// of the whole vector (§7.4.1).
if self.packed_elem_width_of(hier).is_some() { return None; }
```

but `packed_elem_width_of` ends in `.filter(|&w| w > 1)`, which drops a
unit-width element, so the guard never fired. `[N-1:0][1:0]` was correct
because its width passed the filter and the guard bailed to the element-aware
interpreter path.

The filter is right for the function's other callers: on a single-index select,
element `i` of `[N-1:0][0:0]` is physically bit `i`, so `v[i]` may legitimately
fuse as a bit-select. That is why `assign v[p] = x` always worked. It is wrong
only at this guard, where the question is shape, not width.

### Fix

`is_packed_multi_dim(hier)` is the same lookup without the width filter. With
the guard asking that instead, `v[i][0]` bails to the element-aware interpreter
path, the route `[1:0]` already took. This is a sibling of `0f89500`, which
fixed the same mis-fusion for `elem_w > 1` and left `[0:0]` behind.

Two dead ends, so nobody retraces them. `FusedGate` is not involved:
`try_resolve_bit_ref`'s `Index` arm also requires an `Ident` base, and
`v[3][0]` has an `Index` base. Elaboration is fine: `packed_inner_elem_width`
has no `> 1` guard, so `[4:0][0:0]` registers
`packed_signal_elem_widths["v"] = 1` correctly.

FlooNoC's router internals use `[NumRoutes-1:0][NumVirtChannels-1:0]` with
`NumVirtChannels == 1` throughout, so this hits the DUT everywhere and cannot
be worked around in the testbench. On our bench accepted flits went from 10 to
21: the driver now handshakes on all five ports instead of only port 0.

## 2. `$size(a.M)` on a class member array returns 32

`src/compiler/simulator.rs`

```
                     before   Verilator
$size(M)   in-method    3         3      correct
$size(a.M) qualified   32         3      wrong
$size(a.A) qualified   32         3      wrong (ascending too)
```

`$left` and `$high` report 31 the same way. The operand's bit width comes back
in place of the declared element count.

### Cause

The §20.7 query path only resolved a plain `ExprKind::Ident`. The qualified
spelling parses either as a `MemberAccess`, which never entered that arm, or as
a two-segment `Ident` resolving to a name with no registered shape. Either way
no bounds were found and the query fell through to the final fallback, which
evaluates the operand and reports `v.width`.

The shape was available all along: `new` registers every fixed-size array
property under `<handle>#<member>` in the `arrays` / `arrays_2d` / `arrays_nd`
tables, which is why `foreach (a.M[i])` already iterates correctly. Only the
query path missed the mapping.

### Fix

`handle_member_array_name` resolves the operand to that store, after which the
existing declared-bounds path answers the query unchanged. It accepts both
spellings, the pair `assign_class_collection_pattern` already handles for
whole-array writes, and walks the runtime class ancestry from
`heap[handle].class_name` so inherited members resolve.

It returns `None` unless the class really declares the member as a fixed-size
array, so module hierarchical paths and packed non-array members keep their
behaviour: `$size(c.big)` on a `bit [88:0]` field still reports 89.

### Scope

`$left` and `$right` on a descending member (`int M[2:0]`) still report `0` and
`2` where Verilator reports `2` and `0`. That is separate and pre-existing:
`array_properties` normalizes bounds with `(l.min(r), l.max(r))` in
`elaborate.rs`, and the in-method `$left(M)` is wrong the same way. Fixing it
ripples through every consumer of those bounds, so this PR only makes the
qualified form agree with the in-method one. The tests assert direction on the
ascending case only.

## Tests

Six tests in the existing group layout. Expected values are Verilator 5.050's.

`tests/gates/cont_assign_packed2d_unit_inner.rs`: explicit indices with no
generate loop, which pins that the genvar is not the factor; the generate form
next to the three spellings that already worked, so a regression there shows
the widened guard pulled a working form off its fast path; the procedural
write, which takes a different path.

`tests/classes/array_query_handle_qualified_member.rs`: qualified queries
against the in-method ones, with a per-dimension query on a 2-D member and
`foreach` pinned alongside; `this.M` and an inherited member through a handle;
a packed non-array member keeping its bit-width answer.

## Verification

`cargo test --release --features libffi/system`: 2067 passed, 8 failed. All 8
are `cc failed` in `dpi_integration_tests` and `dpi_unit_scope`, the
pre-existing macOS `-shared` VPI-symbol linker issue that fails during C
compilation before xezim runs. They fail on `main` too.

Both repros were cross-checked against Verilator 5.050. Built on `main` at
`8a20974` against the stock pinned core `22a170e`.

## Known follow-up: the width proxy should be a shape predicate

`is_packed_multi_dim` is the third place that asks about packed shape by
testing an element width. `try_resolve_bit_ref` in `simulator.rs` and
`sel_base_needs_ast` in `bytecode.rs` do the same, and `sel_base_needs_ast`
applies `.filter(|&w| w > 1)` on top of `packed_elem_width_of`, whose body
already ends in that filter.

Width cannot answer the question. Whether an index chain may fuse as a bit
reference depends on how many packed dimensions it consumes and what width is
left over. `v[i]` on `[N-1:0][0:0]` fuses correctly today only because the
filter happens to drop it, while `v[i][0]` on that signal and on `[N-1:0][1:0]`
both resolve to a single bit at a computable flat offset yet bail to the
interpreter.

One predicate over declared dimensions and index depth would replace all of it,
and most of the machinery exists: `packed_full_dims` carries the dimensions and
reaches both layers, `chain_root_dims` / `packed_elem_lsb` /
`emit_packed_slot_index` and the chained-select emitter already do the slot
arithmetic including ascending and non-zero-based outer dimensions, and
`elem_chain_needs_ast` counts index depth against `dims.get(layers)`. It stays
out of this PR because it touches roughly ten sites per file, needs its own
benchmark (these are the fast paths `0f89500` protects), and would widen fusion
on a path that currently returns correct answers.


aionhw (approved):
Both fixes reproduced on current `main` before review, and both are gone with the branch:

| case | main | this PR |
|---|---|---|
| `assign v[i][0]` on `logic [4:0][0:0] v` | `xxxx1` | `01101` |
| `$size(a.M)` / `$size(a.A)` on class members | 32 / 32 | 3 / 3 (`$size(a.big)` still 89) |

The diagnosis in the description matches the code: `compile_blocking_target`'s guard asked a width question (`packed_elem_width_of`, which filters unit-width elements) where the question is shape, so `v[i][0]` fell to `flattened_outer_const_signal_id` and every element wrote bit 0. `is_packed_multi_dim` is the right predicate there. `handle_member_array_name` resolving `a.M` / `this.M` to the `<handle>#M` store, gated on the class chain really declaring a fixed-size array, keeps packed members and module paths on their existing paths.

Verified here on Linux against the current `main` (merged into the branch locally):

- the six new tests plus the full `gates` (95) and `classes` (329) groups pass;
- the full default suite passes on the merged tree;
- the C906 hello gate is bit-exact, so the widened guard does not disturb the compiled RTL paths;
- the six AVIP UVM benches are at baseline.

One note for the merge: the branch's last merge of `main` predates 0.10.5. On the PR head as pushed, `tests/hierarchy/bind_grid_module_name_refs.rs` overflows the stack, but that is `main`'s same-name `ref` formal defect fixed in 17d004f (the parametrised `bind` in that test only started being applied today, bf004b5 in core), not anything in this branch; it passes once `main` is merged in. A rebase or merge before merging keeps CI green.

The follow-up you list (a shape predicate replacing the three width proxies, `sel_base_needs_ast` included) is worth doing as its own PR.

Approve.

aionhw (approved):
Both fixes reproduced on current `main` before review, and both are gone with the branch:

| case | main | this PR |
|---|---|---|
| `assign v[i][0]` on `logic [4:0][0:0] v` | `xxxx1` | `01101` |
| `$size(a.M)` / `$size(a.A)` on class members | 32 / 32 | 3 / 3 (`$size(a.big)` still 89) |

The diagnosis in the description matches the code: `compile_blocking_target`'s guard asked a width question (`packed_elem_width_of`, which filters unit-width elements) where the question is shape, so `v[i][0]` fell to `flattened_outer_const_signal_id` and every element wrote bit 0. `is_packed_multi_dim` is the right predicate there. `handle_member_array_name` resolving `a.M` / `this.M` to the `<handle>#M` store, gated on the class chain really declaring a fixed-size array, keeps packed members and module paths on their existing paths.

Verified here on Linux against the current `main` (merged into the branch locally):

- the six new tests plus the full `gates` (95) and `classes` (329) groups pass;
- the full default suite passes on the merged tree;
- the C906 hello gate is bit-exact, so the widened guard does not disturb the compiled RTL paths;
- the six AVIP UVM benches are at baseline.

One note for the merge: the branch's last merge of `main` predates 0.10.5. On the PR head as pushed, `tests/hierarchy/bind_grid_module_name_refs.rs` overflows the stack, but that is `main`'s same-name `ref` formal defect fixed in 17d004f (the parametrised `bind` in that test only started being applied today, bf004b5 in core), not anything in this branch; it passes once `main` is merged in. A rebase or merge before merging keeps CI green.

The follow-up you list (a shape predicate replacing the three width proxies, `sel_base_needs_ast` included) is worth doing as its own PR.

Approve.

PR #157, opened 2026-09-07
Add VPI support for ports and port directions

oscargus:


aionhw (changes requested):
Thanks for taking this on — port objects are a real gap in the VPI layer. I built the branch and probed it with a small VPI module, and as submitted the feature never fires, so this needs another pass before it can go in.

**1. The direction map is always empty.** `vpi_port_directions` is built at simulator.rs ~8077 from `module.signals`, but `Simulator::new` has already freed that map by then (`module.signals = HashMap::default()` at ~7980, and earlier at ~7418, once the indexed `signal_table` exists). A trace at the construction site prints `n=0 keys=[] total_signals=0`. So `vpi_type_of` never returns `vpiPort`, `vpiDirection` is `vpiNoDirection` for every port, and `vpi_iterate(vpiPort, ...)` is empty:

```
module top(input logic clk, input logic [3:0] i, output logic [3:0] o, inout wire [1:0] io);
vpiPort(top) = []
vpiReg(top)  = [clk:48/5,i:48/5,internal:48/5,io:48/5,o:48/5]     // type/direction
by_name top.clk type=48 dir=5
```

Build the map before the first `module.signals` clear (the loop at ~7048 still sees the full map, and port signals do carry `direction: Some(..)` from elaboration), or read direction from the elaborated `Signal` while the `signal_table` is being indexed.

**2. Retyping the signal to `vpiPort` is not the object model the standard describes.** In 1800-2017 §37.16 a port is a separate object: `vpi_iterate(vpiPort, module)` yields port objects with `vpiDirection`, `vpiPortIndex`, `vpiSize`, and `vpiHighConn`/`vpiLowConn` to the connected nets; the net or variable behind the port stays a `vpiNet`/`vpiReg`. Once the map is populated, the current `vpi_type_of` change would make every port signal disappear from `vpiNet`/`vpiReg` iteration — `tests/dpi/vpi_object_model.c` line 98 expects `vpi_iterate(vpiReg, sub)` to list `clk,i,o` and would start failing (it passes today only because the map is empty). Suggest a `VpiKind::Port` handle produced by the `vpiPort` iteration arm, carrying the direction and pointing at the underlying signal id, and leave `vpi_type_of` alone.

**3. Sub-instance ports.** The inliner recreates instance-scope signals with `direction: None`, so even with fix 1 only the top module's ports would be found. If sub-module ports are in scope for this PR, the direction has to be recorded per instance during inlining (the `port_order`/port list of the instantiated module is available there).

**4. No test.** Please add a `tests/dpi/vpi_ports.c` + `.sv` pair driven from `tests/strings/dpi_integration_tests.rs` like `vpi_object_model_test`: iterate `vpiPort` on the top and on a sub-instance, check name, direction (input/output/inout), size, and that `vpiReg`/`vpiNet` iteration is unchanged.

Small: `vpiSigned 65` moved below the `vpi_time` block — harmless, but keep it where it was to make the header diff about the new constants only. The new constant values (vpiPort 44, vpiPortBit 45, vpiDirection 20, direction values 1–5) match the standard.

oscargus:
Sorry about thar. I had an earlier branch that I carefully tested, but then accidentally ran cargo fmt on... It looked quite similar this time, but should have been more careful.

aionhw:
Merged and reworked into main in 592713f. What changed relative to the branch: the direction map is built in `Simulator::new` before `module.signals` is freed; `vpi_iterate(vpiPort, scope)` yields dedicated port objects (`VpiKind::Port`, type `vpiPort`, with `vpiName`/`vpiFullName`/`vpiDirection`/`vpiSize` and value reads through the connected signal) for the top module and for sub-instances; `vpi_type_of` is unchanged so nets and variables keep their `vpiNet`/`vpiReg` types, and a signal handle answers `vpiDirection` with its port's direction. Test: tests/dpi/vpi_ports.{c,sv} driven by `vpi_ports_test`. Thanks for starting this.

PR #158, opened 2026-09-07
Five write-path and scheduling fixes

chenbhil:
Five commits, one per defect. Each stands alone and can be reverted on its
own. I grouped them because they all turned up the same way, bringing up a FlooNoC router UVM smoke bench that Verilator 5.050 passes, and because four of the five are the same kind of defect: a procedural write that resolved to the wrong storage, or to no storage at all, and disappeared without a diagnostic.

| # | Commit | Defect |
|---|---|---|
| 1 | `Discard a bit-select write whose index is x or z` | §11.5.1: an x/z index wrote bit 0 from inside an always block |
| 2 | `Don't retire the time-0 comb seed before the graph exists` | a constant-RHS continuous assign never evaluated in any UVM testbench |
| 3 | `Infer the element width of an indexed virtual-interface lvalue` | `vif.d[i] <= val` truncated to 1 bit |
| 4 | `Resolve a NESTED packed-struct member lvalue to its registered base` | `f.hdr.d = v` dropped; only depth-1 members resolved |
| 5 | `Write class unpacked-struct array elements to the store their reads use` | element writes and element reads used two different stores |

---

## 1. A bit-select write with an x/z index is not discarded

§11.5.1 says a bit-select write at an unknown or out-of-range index modifies
no bits. Out-of-range was already correct, because `Value::set_bit` drops it.
An x/z index wrote bit 0.

```systemverilog
logic [2:0] sel;        // never assigned, stays x
logic [4:0] a;
always_comb begin
  a = '0;
  a[sel] = 1'b1;        // before: 00001    LRM: 00000
end
```

`Value::to_u64` returns `Some(val_bits & !xz_bits)`. It masks x bits to zero
and never returns `None`, so the `to_u64().unwrap_or(0)` behind every dynamic
bit-store had no way to tell an unknown index from a genuine 0. An `initial`
block and a task were already correct, which is why this only showed up from
inside an always block.

A shared `dyn_bit_index` now returns `None` on `has_xz()`. It is used at all
five interpreter sites: the sequential, view and parallel executors, in both
the blocking and NBA forms. The JIT and AOT backends reach the same stores
through the `blk_range` and `nba_bit` bridges and were passing only the
index's value plane, so the x plane never arrived. Both now hand the bridge a
deliberately out-of-range index when the x plane is non-zero, which those
bridges already drop. The ABI is unchanged.

One caveat on the test: Verilator cannot referee this. It is 2-state here, so
it resolves the index to 0, performs the write, and prints `00001` for the
broken cases. The expected values in the test come from the LRM.

## 2. A constant-RHS continuous assign never drives in a UVM testbench

`assign a[i] = '0;` never evaluated and its net kept `x` for the whole run, in
any design that also has a parameterized class with a `type_id` typedef. That
is to say, in every UVM testbench.

An entry whose RHS is constant has an empty read set, so nothing ever marks it
dirty. It reaches the settle worklist through exactly one path: the one-shot
`comb_time0_idx` seed in `settle_combinatorial_inner`, which the first settle
at time 0 consumes before latching `comb_time0_fired`.

`settle_after_proc_write` runs a settle after every procedural blocking write.
A parameterized class with a `type_id` typedef has its statics initialized in
the `registry spec static init` elaboration phase, which runs before `build
combinational entries`. So that settle finds an empty entry list. It seeds
nothing and retires the one-shot anyway. The real time-0 settle then skips
`comb_time0_idx`, and every constant-driven net in the design goes
unevaluated: 251 of them in the bench.

The fix is to require a non-empty entry list before the one-shot is consumed,
in both `settle_combinatorial_inner` and `settle_combinatorial_bsp`.

The step that cracked it was odd enough to be worth passing on. A 50-line
non-UVM harness instantiating the DUT directly does not reproduce. Adding
`uvm_pkg.sv` to the compile alone, with UVM still never instantiated, does. A
design that changes because an unused package got compiled is a global table
or a global one-shot, never the RTL.

## 3. An NBA to a packed-array element through a virtual interface keeps only the LSB

```systemverilog
task nba(int i, logic [15:0] val); vif.d[i] <= val; endtask   // wrote 0x0001
```

Exactly one combination breaks, NBA plus virtual interface. The blocking form
is fine, a module-scope array is fine, and a hierarchical interface reference
(`i0.d[k] <= ...`) is fine.

`infer_lhs_width`'s `ExprKind::Index` arm walks the chained index nodes down
to a root and then asks the dimension maps how wide the selected element is,
but it accepted only an `Ident` root. `vif.d[i]` has a `MemberAccess` root, so
the whole block was skipped and the lvalue fell through to the trailing `1`.

That width is applied at schedule time, in the `resize_for_assign` that runs
before the entry is queued. The value was already one bit wide before anything
at the commit site could see it. Hence the blocking form being unaffected: it
never consults `infer_lhs_width`. Hence also an `Ident`-rooted array always
being right.

A new helper, `vif_member_flat_name`, resolves `vif.member` and
`this.vif.member` to the bound interface signal's flat name, which is what the
packed-dimension maps are keyed by. It is the same lookup `resolve_nba_target`
already does. The width logic itself is untouched; the arm just reaches it for
one more root shape.

I spent a while on the wrong patch first, which may save someone else the
trip. My first attempt went into `assign_value`'s §25.8 vif arm, which does
write bit `i` unconditionally and looks exactly like the culprit. It changed
nothing, because that arm is not on this path. The truncation only showed up
after following the NBA queue back to its schedule site.

## 4. A nested packed-struct member write is dropped when the lvalue is a MemberAccess

`f.hdr.d = mk_id();` vanished. `f.pl = 64'hface...;` in the same function
landed.

`packed_struct_fields[<decl>]` is flattened. Every nested member is a key in
its own right (`"hdr.d.x"` under `"f"`), and there is no entry for an
intermediate member on its own. The `MemberAccess` arm of `assign_value_inner`
split the lvalue at exactly one dot, so `f.hdr.d` asked for
`packed_struct_fields["f.hdr"]`, got `None`, declined, and the write fell
through every remaining arm. Only a depth-1 member ever resolved.

A new `packed_struct_member_slice` walks the split points longest base first,
which is the same candidate walk `resolve_packed_struct_target` already does
for assignment patterns. That asymmetry explains the whole variant table:
`f.hdr.d = '{x:2, y:1}` worked because the pattern path had the walk, and
`f.hdr.d = mk_id()` failed because the value path did not. The first candidate
tried is the old behaviour, so existing resolutions are unchanged.

Two things about my original framing were wrong, and the commit records them
so nobody re-derives them. This is not about subroutine-local storage: a task
writing a module-scope signal fails identically, and the real discriminator is
the lvalue shape. And reads were never broken, at any depth.

Known gap, left in deliberately: a class property base still drops the write.
The member now resolves, but `get_local_or_signal` reaches only the frame and
the signal table, not heap properties, so the arm declines at the storage step
instead of the lookup step. That wants a separate storage-routing patch.

## 5. Class unpacked-struct array elements: writes and reads used two stores

```systemverilog
class C;
  u_t A [3];
  function void go();
    A = '{ '{4,2'd2}, '{5,2'd1}, '{6,2'd3} };   // stored, then unreachable
    $display("%0d", A[0].s);                    // 0
  endfunction
endclass
```

`queue_elem_struct` matched a class property that is a fixed array as though
it were a queue, so a whole-element write took the instance-scoped signal
route that queues legitimately use. A fixed array's reads resolve through
`class_unpacked_leaf` into the instance property map. Printing the two keys
side by side is what caught it:

```
SIGWRITE "1#BELEM[0].s" = 4
read consults heap[1].properties["BELEM[0].s"]
```

One rule at three sites: ask the read resolver where the leaves live, and
write there. The sites are the whole-element write (both RHS forms), the
whole-array pattern, hooked above the `spread` block that was consuming it,
and `localparam`/`const` construction. The guard is a question put to the read
path rather than a heuristic, so a genuine queue, dynamic array or associative
element is not claimed and keeps its signal-name copy, which UVM's
`uvm_hdl_path_concat::add_slice` depends on.

Three earlier framings were wrong here too, and the commit records them:
"unpacked vs packed element type" confounded element type with storage class,
"a `localparam` array reads back all-zero" had the wrong factor since
`localparam` is not involved, and "`coll_assoc_struct_elem` returns a hard
`Value::zero(w)`" was simply not what happened, since that probe never fired
and the read path is correct throughout.

One thing this does not fix. A scalar unpacked-struct property written with a
pattern (`u_t S; S = '{9,2'd1};`) still reads back 0. It has no array element
and reaches none of the three sites. I built the parent commit to check, and
it fails there too, so this PR is not the cause.

---

## Tests

Each commit carries its own, placed next to the existing siblings:

- `tests/types/bit_write_unknown_index_discarded.rs`, beside the read-side
  `array_read_unknown_index_is_x.rs`. It covers the x-index write in
  `always_comb`, `always_ff` and `initial`, the index forms that already
  worked so the fix cannot suppress a real write, and the arbiter shape
  itself.
- `tests/gates/const_cont_assign_time0_seed.rs`
- `tests/hierarchy/vif_packed_array_elem_nba.rs`
- `tests/types/nested_packed_member_lvalue.rs`
- `tests/classes/class_unpacked_struct_array_store.rs`

The last two are new in this submission, and I checked both the way round that
actually matters: each one fails on the parent commit of its fix and passes
with it, while its control assertions (reads unaffected, the pattern-RHS
control, the forms that already worked) pass in both states.

## Verification

`cargo test --release --features libffi/system --no-fail-fast`, on `main` at
`a00eb8b` against the stock pinned core, since this PR takes no core change:
2538 passed, 11 failed. All 11 are the pre-existing macOS `cc -shared`
DPI/VPI linker failures. They fail in `cc` before xezim runs, and the same
set fails on stock `main`.

End to end, with these five plus the companion core PR, the FlooNoC router UVM
smoke bench goes from a t=50000 watchdog timeout with 0 flits delivered to
TEST PASSED: 25 accepted, 25 delivered, 25 matched, 0 mismatches, 0 UVM_ERROR,
matching the Verilator reference. Its three fault-injection self-tests still
fail as designed, so the checkers are not vacuous.

https://claude.ai/code/session_01VNpHRwfUC38QkpLUmW3PJd

### DIFF
--- a/src/compiler/aot.rs
+++ b/src/compiler/aot.rs
@@ -1014,7 +1014,15 @@ fn emit_insn_rust(
         }
         BlockingAssignBitDyn(sig, idx_r, val) => {
             let sig = sref(*sig);
-            let _ = writeln!(w, "(br().blk_range)(sim, {sig}, r{idx_r}v, r{idx_r}v, r{val}v, r{val}x);");
+            // §11.5.1: an x/z index modifies no bits. The x plane never
+            // reached the bridge and the value plane has x bits masked to
+            // zero, so an unknown index arrived as bit 0. Hand the bridge a
+            // deliberately out-of-range index instead; it already drops those.
+            let _ = writeln!(
+                w,
+                "let bi{idx_r} = if r{idx_r}x != 0 {{ u64::MAX }} else {{ r{idx_r}v }};\n\
+                 (br().blk_range)(sim, {sig}, bi{idx_r}, bi{idx_r}, r{val}v, r{val}x);"
+            );
         }
         NbaAssignConst(sig, v, width) => {
             let sig = sref(*sig);
@@ -1037,7 +1045,12 @@ fn emit_insn_rust(
         }
         NbaAssignBitDyn(sig, idx_r, val) => {
             let sig = sref(*sig);
-            let _ = writeln!(w, "(br().nba_bit)(sim, {sig}, r{idx_r}v, r{val}v, r{val}x);");
+            // §11.5.1, as for the blocking form above.
+            let _ = writeln!(
+                w,
+                "let bi{idx_r} = if r{idx_r}x != 0 {{ u64::MAX }} else {{ r{idx_r}v }};\n\
+                 (br().nba_bit)(sim, {sig}, bi{idx_r}, r{val}v, r{val}x);"
+            );
         }
         LoadArrayElem(d, arr, idx_reg) => {
             let ArrayOperand::Dense { first_id, lo, hi, .. } = arr.as_ref() else {
--- a/src/compiler/jit.rs
+++ b/src/compiler/jit.rs
@@ -1953,9 +1953,7 @@ mod enabled {
             BlockingAssignBitDyn(sig_id, idx_reg, val_reg) => {
                 let (v, x) = ld2(builder, pointer_type, regs, xz, *val_reg);
                 let id = builder.ins().iconst(types::I32, *sig_id as i64);
-                let idx = builder
-                    .ins()
-                    .stack_load(pointer_type, types::I64, regs[*idx_reg as usize], 0);
+                let idx = dyn_bit_index_or_oob(builder, pointer_type, regs, xz, *idx_reg);
                 // A 1-bit write at [idx:idx] — same semantics, same bridge.
                 builder
                     .ins()
@@ -2099,9 +2097,7 @@ mod enabled {
             NbaAssignBitDyn(sig_id, idx_reg, val_reg) => {
                 let (v, x) = ld2(builder, pointer_type, regs, xz, *val_reg);
                 let id = builder.ins().iconst(types::I32, *sig_id as i64);
-                let idx = builder
-                    .ins()
-                    .stack_load(pointer_type, types::I64, regs[*idx_reg as usize], 0);
+                let idx = dyn_bit_index_or_oob(builder, pointer_type, regs, xz, *idx_reg);
                 builder.ins().call(nba_bit_ref, &[sim_ptr, id, idx, v, x]);
             }
             // NbaAssignRangeDyn / NbaAssignBitDyn still left out — they
@@ -2908,6 +2904,26 @@ mod enabled {
     }
 
     /// Load both planes of a VM register.
+    /// §11.5.1: the index operand of a dynamic bit-store, with an x/z index
+    /// turned into a deliberately OUT-OF-RANGE one. The store bridges already
+    /// drop an out-of-range index, so this makes "unknown index writes
+    /// nothing" fall out of the path that is already there. The x plane never
+    /// reached the bridge, and the value plane has x bits masked to zero, so
+    /// an unknown index arrived as a perfectly ordinary bit 0.
+    fn dyn_bit_index_or_oob(
+        builder: &mut FunctionBuilder,
+        pointer_type: Type,
+        regs: &[StackSlot],
+        xz: &[StackSlot],
+        r: u16,
+    ) -> Value {
+        let (iv, ix) = ld2(builder, pointer_type, regs, xz, r);
+        let zero = builder.ins().iconst(types::I64, 0);
+        let oob = builder.ins().iconst(types::I64, -1);
+        let unknown = builder.ins().icmp(IntCC::NotEqual, ix, zero);
+        builder.ins().select(unknown, oob, iv)
+    }
+
     fn ld2(
         builder: &mut FunctionBuilder,
         pointer_type: Type,
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -42062,6 +42062,56 @@ impl Simulator {
         false
     }
 
+    /// LRM §25.8 — resolve `vif.member` (or `this.vif.member`) to the FLAT
+    /// name of the bound interface signal, `"<iface_instance>.<member>"`.
+    ///
+    /// `resolve_nba_target` asks the same question but only ever returns a
+    /// registered signal id, which is no use to a caller that needs the NAME
+    /// — to look up the member's packed dimensions, say. The binding lives
+    /// either in `local_iface_aliases` (a vif passed as a task formal) or in
+    /// `virtual_iface_bindings` keyed by the current `this`; both lookups are
+    /// the ones that arm already does.
+    fn vif_member_flat_name(&self, expr: &Expression) -> Option<String> {
+        let ExprKind::MemberAccess { expr: base, member } = &expr.kind else {
+            return None;
+        };
+        let prop = match &base.kind {
+            ExprKind::Ident(hier) if hier.path.len() == 1 => {
+                hier.path[0].name.name.as_str()
+            }
+            ExprKind::MemberAccess {
+                expr: inner,
+                member: vprop,
+            } if matches!(&inner.kind, ExprKind::This) => vprop.name.as_str(),
+            _ => return None,
+        };
+        if let Some(bound) = self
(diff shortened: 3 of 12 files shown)


aionhw (changes requested):
Thanks — this is a careful set, and the write-ups (especially the "unused package changes the design" step in fix 2 and the wrong-patch note in fix 3) are exactly the kind of record worth keeping. I trialled the branch merged onto current main (one trivial conflict in `tests/hierarchy.rs`, both sides appended a `mod` line): the five new tests pass, also under `XEZIM_JIT=1`; default and JIT suites clean; the c906 hello gate is bit-exact; the six AVIP baselines are unchanged; the axi4 UVM output is identical.

One thing blocks it, and it is measurable rather than a matter of taste.

**Fix 5 costs +1.3% instructions on the axi4 UVM benchmark.** Same base, same binary apart from the commit:

| tree | axi4 insns |
|---|---|
| main | 33.06G |
| PR (all five) | 33.51G |
| PR with fix 5 reverted | 33.08G |

So fixes 1–4 are neutral and fix 5 is the whole increase. The cause is the second hook site: `if self.queue_pop_call(rvalue).is_none() && self.class_unpacked_elem_is_heap_owned(lvalue)` runs for every blocking assignment that reaches that arm whenever any class object exists, which is every assignment in a UVM run. `class_unpacked_elem_is_heap_owned` calls `class_unpacked_array_prop_of` (receiver lookup, `class_prop_struct`, `spreads_member_wise`), then builds an `append_member_expr` probe and runs `class_unpacked_leaf` — and the caller then calls `class_unpacked_array_prop_of` a second time to get `su`. All of that for an lvalue shape (`A[i]` on a class property that is a fixed array of unpacked structs) that is rare.

Suggested shape: decide from the lvalue syntax first and only then resolve. The write in question always has an `Index` (or `Ident` with a select) whose root is a class property; precompute per class the set of property names that are fixed arrays of unpacked structs (`property_types` is available at class registration), and check the lvalue's root identifier against that set before calling any resolver. Also return `su` from the ownership check so it is computed once. With that gate the UVM path should pay one name compare per assignment. Please re-measure axi4 (`perf stat -e instructions` on the axi4 AVIP base test, A/B against main) and quote the numbers in the commit.

Notes, not blockers:

- Fix 2: the guard is added in `settle_combinatorial_inner` and `settle_combinatorial_bsp`, but the opt-in per-partition variant (`settle_combinatorial_perlp`, `XEZIM_PERLP_SETTLE`) still latches `comb_time0_fired` on an empty entry list (~46152). Same one-line guard there keeps the three in step.
- Fix 1: the JIT and AOT bridges pass an all-ones index and rely on `Value::set_bit` dropping it; that works (the NBA path enqueues a no-op entry in that case). Fine as is; a one-line comment at the bridge saying the drop is intentional would save the next reader the trace.
- Fix 4's known gap (class-property base) is stated honestly; good.
- The tests cite the reference simulator's values, which is the convention here — thanks for keeping the reference unnamed.

Once fix 5 is gated and re-measured I expect this to be a straightforward approve.

aionhw:
Merged into main via a merge commit (6732ac2) with the fix-5 gate applied in the follow-up commit 592713f: `lvalue_root_is_unpacked_struct_prop` checks the lvalue's root identifier against a once-built set of class property names whose (typedef-resolved) type is an unpacked struct, and skips a bare single-identifier lvalue outright, before any resolver runs. axi4 instructions: 33.51G on the PR as submitted → 33.12G with the gate (main before the merge 33.06G; fixes 1–4 are neutral). Your five tests, both suites, the hello gate and the six AVIP baselines pass on the merged tree. Thanks for the fixes.

Issue #161, opened 2026-09-10
Time literals in constant expressions are scaled to a fixed 1 ns instead of the declaring module timescale (sub-ns delays round to zero)


vrajeshprakhya:
A time literal is scaled to the current time unit when it is evaluated at run time, but the scaling is dropped when the same literal is evaluated at elaboration. `#100ns` is correct; `localparam realtime T = 100ns` is not.

IEEE 1800-2017 5.8:

> Time is written in integer or fixed-point format, followed without a space by a time unit (`fs ps ns us ms s`). ... The time literal is interpreted as a realtime value scaled to the current time unit and rounded to the current time precision.

There is no exemption for constant expressions, and 6.20.2 does not carve one out for parameter declarations.

### Repro

<details>
<summary><code>timelit.test.sv</code> — self-checking, prints TIMELIT: PASS or FAIL</summary>

```systemverilog
`timescale 1fs/1fs
// The time unit is 1 fs, so every value printed must be 100000000 (100 ns).
//
// 1 fs is chosen only because it makes the two behaviours differ. Under
// `timescale 1ns/1ns` a tool that drops the suffix prints 100 and is
// accidentally right, which is what makes this quiet in normal use.
module timelit #(parameter realtime P_PORT = 100ns) ();

  localparam realtime P_LOCAL = 100ns;
  const     realtime  C_CONST = 100ns;
  realtime            v_init  = 100ns;   // static (elaboration) initialiser
  realtime            v_proc;

  localparam realtime EXP = 100000000.0;
  int errors = 0;

  task automatic chk(string what, realtime got);
    if (got == EXP) $display("  ok    %-24s %0.0f", what, got);
    else begin
      $display("  WRONG %-24s %0.0f   expected %0.0f", what, got, EXP);
      errors++;
    end
  endtask

  initial begin
    realtime a_auto = 100ns;             // automatic initialiser
    realtime t0;

    v_proc = 100ns;

    $display("timescale 1fs/1fs -- every value below must be %0.0f", EXP);
    $display(" elaboration-time evaluation:");
    chk("parameter port",      P_PORT);
    chk("localparam",          P_LOCAL);
    chk("const",               C_CONST);
    chk("static var init",     v_init);

    $display(" run-time evaluation:");
    chk("automatic var init",  a_auto);
    chk("procedural assign",   v_proc);

    t0 = $realtime; #100ns;     chk("delay #100ns",    $realtime - t0);
[...]

PR #42, opened 2026-07-21
Add cross-platform install scripts (Linux/macOS + Windows) and Quick Install section to README

opensource-elearning:
### Summary

  Adds production-grade one-liner installers for all platforms and a Quick Install section
  at the top of the README.

  ### Files Added

  - **scripts/install.sh** — Unified installer for Linux & macOS
    - Pre-flight checks (git, curl/wget, Xcode CLI tools)
    - Auto-installs Rust via rustup if missing
    - Auto-detects latest release tag from git (or `XEZIM_TAG` env var to pin)
    - Clones xezim-core + xezim side-by-side, builds release
    - Auto-installs to PATH (symlink to /usr/local/bin → ~/.local/bin → shell config)
    - Smoke test after build
    - Verbose emoji output (✅ ⚠️ ➡️ ❌)
    - Offline support (skips clone + fetch if repos already exist)

  - **scripts/install.ps1** — Windows PowerShell installer
    - Installs Rust via rustup-init.exe if missing
    - Installs Git via winget → GitHub API (dynamic latest version)
    - Same tag detection, build, and PATH setup as install.sh
    - Save/restore pattern for RUSTFLAGS to avoid leaking

  ### Files Modified

  - **README.md** — Added Quick Install section with Linux/macOS and Windows
    one-liner commands (with `XEZIM_TAG` for version pinning)

  ### Usage

  ```bash
  # Linux & macOS
  curl -fsSL https://raw.githubusercontent.com/aionhw/xezim/main/scripts/install.sh | sh
  XEZIM_TAG=v0.9.6 curl -fsSL https://raw.githubusercontent.com/aionhw/xezim/main/scripts/install.sh | sh

  # Windows (PowerShell)
  irm https://raw.githubusercontent.com/aionhw/xezim/main/scripts/install.ps1 | iex
  $env:XEZIM_TAG='v0.9.6'; irm https://raw.githubusercontent.com/aionhw/xezim/main/scripts/install.ps1 | iex

PR #162, opened 2026-09-10
Regression tests: time-literal scaling in declaration initialisers (#161)

vrajeshprakhya:
Regression tests for aionhw/xezim#161. The fix itself is in the companion xezim-core PR; this pins the behaviour where it is observable, per AGENTS.md.

**This PR needs the core `rev` bump to go green** — the tests fail against the currently pinned core, which is the point of them.

## What is covered

`tests/scheduling/decl_init_time_literal_scaling.rs`, five cases:

| test | pins |
|---|---|
| `declaration_initialisers_scale_like_procedural_code` | all four elaboration-time forms (parameter port, `localparam`, `const`, static var init) agree with the five run-time forms under `1fs/1fs` |
| `sub_nanosecond_and_micro_second_literals_scale_by_the_unit` | `300ps`→300 and `1us`→1000000 at a 1 ps unit — the cases that distinguish "scaled to the wrong unit" from "suffix discarded" |
| `a_sub_nanosecond_constant_delay_is_not_rounded_away` | the constant SPENT as a delay, which is the shape that failed in the field; it measured `0.000 ps` before |
| `the_same_constant_scales_with_each_modules_own_timeunit` | one body at `1ns`/`1ps`/`1fs`, i.e. the `1ns / timeunit` ratio the bug was |
| `identity_at_the_default_nanosecond_timeunit` | the default-timeunit case is unchanged |

## Verified to fail without the fix

Not assumed — measured, by stashing the core change and rebuilding:

```
test ...identity_at_the_default_nanosecond_timeunit ... ok
test ...declaration_initialisers_scale_like_procedural_code ... FAILED
test ...sub_nanosecond_and_micro_second_literals_scale_by_the_unit ... FAILED
test ...the_same_constant_scales_with_each_modules_own_timeunit ... FAILED
test ...a_sub_nanosecond_constant_delay_is_not_rounded_away ... FAILED

  left: ["NOTE: p300 0", "NOTE: p1u 1000"]
 right: ["NOTE: p300 300", "NOTE: p1u 1000000"]

  left: ["NOTE: elapsed 0"]
 right: ["NOTE: elapsed 300"]
```

4 of 5 fail. The one that passes both ways is the identity test, which is exactly what a compatibility assertion should do — it is there to catch a fix that changes designs at the default timeunit, and it shows this one does not.

## Expected values

Every expected value was cross-checked against Icarus Verilog 14.0 rather than derived only from my reading of §5.8, including the rounding case that makes `300ps` zero at 1 ns precision.

Added to the `scheduling` group (one `#[path]` line), so no new link unit.

### DIFF
--- a/tests/scheduling.rs
+++ b/tests/scheduling.rs
@@ -19,6 +19,8 @@ mod static_local_nba_per_instance;
 mod block_local_decl_ast_fallback;
 #[path = "scheduling/class_method_delay_timeunit.rs"]
 mod class_method_delay_timeunit;
+#[path = "scheduling/decl_init_time_literal_scaling.rs"]
+mod decl_init_time_literal_scaling;
 #[path = "scheduling/fork_children_start_in_spawn_slot.rs"]
 mod fork_children_start_in_spawn_slot;
 #[path = "scheduling/wait_fork_immediate_children.rs"]
--- a/tests/scheduling/decl_init_time_literal_scaling.rs
+++ b/tests/scheduling/decl_init_time_literal_scaling.rs
@@ -1,0 +1,175 @@
+//! §5.8: "The time literal is interpreted as a realtime value scaled to the
+//! current time unit and rounded to the current time precision." There is no
+//! exemption for constant expressions, and §6.20.2 does not carve one out for
+//! parameter declarations.
+//!
+//! The scope-time pass applied that scaling to STATEMENTS only, so procedural
+//! code was correct — `#100ns`, `x = 100ns`, an automatic-variable initialiser
+//! — while a DECLARATION initialiser kept its `NumberLiteral::Time` and fell
+//! through to the fixed-1ns fallback in expression evaluation (`secs * 1e9`).
+//! It was therefore scaled to nanoseconds no matter what the module declared:
+//!
+//! ```text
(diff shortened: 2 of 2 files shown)


vrajeshprakhya:
The fix this pins: aionhw/xezim-core#42. Merge order matters — this PR is red until the core  is bumped.

vrajeshprakhya:
CI on this PR is red for a reason unrelated to it: `main` itself does not compile. `tests/misc/elaboration_shape_fixes.rs` has an unterminated string literal introduced by 90b730c, and GitHub builds this branch merged into `main`, so the failure is inherited rather than caused here.

Evidence:

| commit | CI |
|---|---|
| `ba8d123` — this branch's base | success |
| `90b730c` — current `main` | **failure** |

That parse error is the only error in the failing run.

Fixed separately in #163, kept to that one file rather than folded in here so neither review is muddied.

Once #163 lands, this PR still needs the core `rev` bump from aionhw/xezim-core#42 before it can go green — the five tests fail against the currently pinned core by design, which is the point of them.

aionhw (approved):
Ran the five cases against a build with xezim-core#42 applied: all pass, and they fail without it (the 300ps/1us cases are the right discriminators). This goes green once the core PR lands and the pinned rev is bumped; I will bump the pin with the next push.

PR #43, opened 2026-07-21
CI: add three-stage CI pipeline with lint, install verification, and /fix-lint auto-fix

opensource-elearning:
## Overview

  This PR introduces a complete CI pipeline covering code quality, cross-platform installation, and an automated fix workflow.

  **Depends on:** #42 — the install scripts (`scripts/install.sh`, `scripts/install.ps1`) must be merged first, as the CI workflows reference them directly.

  ## What's Included

  ### 🔍 Code Quality (`ci.yml` — lint job)
  - `cargo fmt --check` enforces Rust style conventions
  - `cargo clippy --no-deps -p xezim` catches code issues in our crate only
  - When lint fails, a PR comment is automatically posted suggesting `/fix-lint` to apply fixes

  ### 🧪 Install Verification (`ci.yml` — install)
  - **Linux**: runs `scripts/install.sh` end-to-end → binary verification → `cargo test --release`
  - Ensures the install scripts work on both platforms on every commit

  ### 🤖 Auto-Fix (`fix-lint.yml`)
  - Triggered by commenting `/fix-lint` on any PR
  - Checks out the PR branch, runs `cargo fmt` + `cargo clippy --fix`, and pushes the formatted result
  - No manual formatting needed — just comment and the bot handles it

  ## Pipeline Structure

  | Job | Runner | Purpose | Blocking |
  |-----|--------|---------|----------|
  | lint | ubuntu-latest | fmt check + clippy | fmt ✅ / clippy ⚪ |
  | install | ubuntu-latest | install script e2e test | ✅ |
  | install-windows | windows-latest | PS1 installer test | ✅ |

  ## How to Use

  1. Push code → CI runs automatically
  2. If lint fails → CI posts a comment suggesting `/fix-lint`
  3. Comment `/fix-lint` on the PR → auto-fix workflow formats code and pushes the fix

@aionhw , FYI...

PR #164, opened 2026-09-10
sim: fix type-parameter class port / static-member / method-dispatch semantics on parameterized classes

Thomasb81:
Simulator fixes on parameterized-class and static-member semantics, verified
byte-for-byte against the reference simulator. Depends on the companion
[`xezim-core` PR](https://github.com/aionhw/xezim-core/pull/43) (this branch
pins `xezim-core` at the same revision the core PR is based on).

**Highlights**

- **Chained type-parameter defaults (§6.25/§8.25).** When a partial
  specialization pads a missing trailing type argument with a parameter default
  that is itself the bare name of an earlier bound parameter
  (`#(type REQ=…, type RSP=REQ)`), resolve it to that earlier parameter's
  already-bound leaf value. This makes a partial `typedef`'s static storage key
  agree with the derived-class read key, so `this_type::` per-spec static live
  in the same cell (e.g. `uvm_sequence_library#(simple_item)`).

- **Type-parameter class port bindings in the callee's own frame.** Resolve the
  specializing arguments of a class formal in the binding site's own frame
  rather than the caller's, so an inherited parameterized base's `extends` chain
  and a typedef-param base carried through a derived class bind correctly
  (§8.25).

- **Static-member / per-spec static semantics (§8.24).** `foreach
  (this_type::m_queue[i])` and bare `this_type::` static reads/writes iterate
  and address the per-specialization static cell even for a dynamic-array /
  queue collection restated in a derived class; bare static-assoc object
  members in instance methods route to the shared store.

- **Method dispatch and name resolution.** Inline blocking static-task calls
  whose receiver is a bare typedef alias; a plain same-named class handle wins
  over a same-named interface; carry a typedef-param base's args through a
  derived `extends`.

- **Tidy output.** Flush simulation stdout before printing end-of-run
  profiling to stderr; sort string-typed queue/array keys lexicographically.

Each fix ships with a focused self-test (see `tests/classes/*`,
`tests/types/*`, `tests/scheduling/*`).


aionhw (changes requested):
Built with xezim-core#43. Results: all ten suite groups green apart from one item below; C906 CoreMark output identical and instruction-neutral (843.95G → 843.98G); C906 hello bit-exact; the six UVM AVIP baselines reproduce exactly. The fixes themselves look right and the tests are focused.

One thing to fix before merge — a flaky new test:

`tests/types/class_type_param_struct_prop.rs` writes the same file (`CARGO_TARGET_TMPDIR/class_type_param_struct_prop/t.sv`) from three tests that run in parallel and each spawns xezim on it. Under the default parallel harness I got `type_param_string_prop_renders_quoted` failing once in three full runs of the types group (`assertion failed` at line 69), and never when run alone or serially — one test truncates/rewrites the file while another's xezim is reading it. Give each test its own file name (or one shared `OnceLock` write) and it is deterministic.

Two notes, not blockers:
- axi4 UVM base test: 31.73G → 32.24G host instructions (+1.6%), output identical. Layout noise on this binary is about ±0.9%, so this may be real; the class-dispatch changes are the likely source if so.
- `env_var_registry` failed on this branch for main's pre-existing ordering, fixed by #166 (merged); a rebase clears it.

PR #165, opened 2026-09-10
Regression tests: whole-struct member expansion (named patterns, nettype drivers)

vrajeshprakhya:
Pin whole-struct member expansion: named patterns, and nettype drivers

Regression tests for aionhw/xezim-core#44. Two files, both in the types
group, both asserting VALUES rather than that something resolved -- each
failing mode produced a struct of zeros with no diagnostic, so a test that
checked only "did the assign happen" would have passed against it. It did
happen, onto members that stayed zero.

struct_named_pattern_continuous_assign.rs -- 4 cases

  * named and ordered forms must agree;
  * named items bind by NAME, not position. The pattern is written in
    reverse member order on purpose: an implementation that mapped named
    items position-wise would pass the first case and fail this one;
  * mixed and nested members, since the expansion recurses;
  * a named pattern reaching a module port.

  The PROCEDURAL form is already covered by struct_named_patterns.rs and
  always worked. This is the CONTINUOUS path, which is a different one --
  expand_whole_struct_continuous_assigns -> emit_struct_member_assigns --
  and the file's name and header say so, so the two are not confused later.

nettype_struct_variable_driver.rs -- 3 cases

  * a nettype net driven from a struct VARIABLE resolves: currents sum to
    2.0 mA and the voltage is the owner's 1.8086, not zero and not an
    average of the two;
  * the assignment-pattern form still resolves -- it always worked, and the
    fix must not disturb it;
  * one element owns a node's voltage while another reads that voltage and
    drives a current back through the SAME port. That is a SPICE node, and
    it is the shape that makes a single port per analog node possible at
    all: without it a real-number model splits every node in two and cannot
    be instantiated against the netlist's own hierarchy.

VERIFIED TO FAIL without the fixes, by stashing them and rebuilding. The
failures are asymmetric, which is the argument for the shape:

    nettype   1 passed, 2 failed   (pattern-driven still passes)
    named     0 passed, 4 failed

In each pair the form that always worked keeps working. A test covering only
that form would have gone green alongside the bug.

Full suite with the fixes in: 2585 passed, 0 failed across all ten groups
plus the core units -- the wide run is deliberate, since the named-pattern
fix changes how every whole-struct continuous assign is expanded.

These need the core `rev` bump from #44 to go green.

Refs: aionhw/xezim-core#44

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_014eLBkW9EERLgZ9jU8TcRSg

### DIFF
--- a/tests/types.rs
+++ b/tests/types.rs
@@ -370,3 +370,7 @@ mod system_function_result_width;
 mod packed_member_select_name_collision;
 #[path = "types/bit_write_unknown_index_discarded.rs"]
 mod bit_write_unknown_index_discarded;
+#[path = "types/nettype_struct_variable_driver.rs"]
+mod nettype_struct_variable_driver;
+#[path = "types/struct_named_pattern_continuous_assign.rs"]
+mod struct_named_pattern_continuous_assign;
--- a/tests/types/nettype_struct_variable_driver.rs
+++ b/tests/types/nettype_struct_variable_driver.rs
@@ -1,0 +1,170 @@
+//! §6.6.7 + §10.3: a user-defined nettype net driven from a struct VARIABLE.
+//!
+//! A nettype whose data type is an unpacked struct is stored as one signal
+//! per member, so a whole-struct value has to be expanded member-wise. The
+//! WRITE side did that already. The driver side did not: drivers were packed
+//! into the resolution-function call as written.
+//!
+//! A driver that is already an assignment pattern survives, because it is
+//! member-wise as authored:
+//!
+//!     assign n = '{i: 3.0e-3, v: 0.0};        // resolved correctly
+//!
+//! A driver naming a struct variable has no whole-struct value to read, so
+//! the resolver received uninitialised members -- nan, or zero once a later
+//! pass defaulted them:
+//!
+//!     nd_t t;  always_comb t = ...;
+//!     assign n = t;                           // members arrived empty
+//!
+//! Both are legal. §10.3 gives `net_assignment ::= net_lvalue = expression`
+//! and an expression may name a variable; the only restriction the clause
+//! puts on a user-defined nettype is on the LEFT-hand side ("shall not
+//! contain any indexing or select"). So the driver is expanded member-wise,
+//! not rejected.
+//!
+//! WHY IT MATTERS beyond the clause. A SPICE node is bidirectional: current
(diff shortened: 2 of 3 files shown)


vrajeshprakhya:
The fixes these pin: aionhw/xezim-core#44.

Merge order matters — this PR stays red until the core `rev` is bumped, which is the point of the tests rather than a problem with them.

vrajeshprakhya:
Correcting this PR's contents and explaining its CI, since neither is what it looks like.

**The branch was wrong.** It was cut from the #163 branch rather than `main`, so it carried that PR's compile fix as well as its own tests — exactly the entanglement I said I was avoiding by keeping them separate, and it also meant this PR was quietly fixing `main` as a side effect. Rebased onto `main`; it now contains three files and nothing else.

**CI here is red for three independent reasons**, and only one belongs to this PR:

| | whose |
|---|---|
| the six new tests failing | this PR's, **by design** — they pin aionhw/xezim-core#44 and cannot pass until the core `rev` is bumped |
| `tests/misc` not compiling | `main`'s, fixed by #163 |
| `env_var_registry` ordering | `main`'s, fixed by #166 |

The first is the point of the tests. From the run:

```
struct_nettype_driven_from_a_pattern_still_resolves ... ok
struct_nettype_driven_from_a_variable_resolves     ... FAILED
  left:  ["NOTE: i=0.0000 v=0.0000 owner=0"]
  right: ["NOTE: i=2.0000 v=1.8086 owner=1"]
```

The pattern-driven case passing while the variable-driven one fails is the asymmetry these were written for: that form always worked, so a test covering only it would have gone green alongside the bug.

Merge order: #163 and #166 to get `main` green, then aionhw/xezim-core#44, then the `rev` bump, then this.

aionhw (approved):
Built against xezim-core#44: all 7 new cases pass; the named-pattern and struct-variable-driver shapes also match the reference simulator on an independent MWE (head without the fix prints zeros for both). misc/scheduling/types groups green apart from main's pre-existing registry ordering (#166). One observation, not a blocker: a_node_can_be_read_and_driven_through_one_port takes ~7 s for a tiny design, so nettype resolution has a cost worth a look some day. Goes green once the core PR lands and the pin is bumped.

PR #166, opened 2026-09-11
Sort XEZIM_CYCLE_CENSUS into place in the ENV_VARS table

vrajeshprakhya:
`main`'s own drift guard fails: `XEZIM_CYCLE_CENSUS` is one line out of order in the `ENV_VARS` table.

`tests/misc/env_var_registry.rs` collects every `"XEZIM_…"` literal in the sources into a `BTreeSet` and compares it against the table, so the printed `--show-env-avail` list cannot rot. Both sides carry the same ~200 entries; only the order differs:

```
table:   … COV_DB, CP_DBG, DBG_ARR, DBG_PARAM, CYCLE_CENSUS, DEAD_CENSUS …
sorted:  … COV_DB, CP_DBG, CYCLE_CENSUS, DBG_ARR, DBG_PARAM, DEAD_CENSUS …
```

`C` sorts before `D`, so the new entry belongs above the two `DBG_*` lines rather than below them.

`XEZIM_CYCLE_CENSUS` was added by 340945e ("Static bounds for constant indexed part-selects; loop census"), which is `main`'s current head — it went in at the wrong position in the same commit that introduced it.

### The fix

One line moved. No other change.

### Verifying it

This is awkward to reproduce, and worth spelling out because the obvious check is inconclusive.

`env_var_registry` lives in the `misc` group, and **`tests/misc` does not compile on `main`** — that is #163, an unterminated string literal in `elaboration_shape_fixes.rs`. So the test never runs in `main`'s CI and its absence there proves nothing either way.

To establish that this failure is real and pre-existing I built `main` plus *only* #163's one-file compile fix, with nothing else applied, and ran the test: it fails, with exactly the ordering diff above.

This PR is **just the one line**. It deliberately does not carry #163's fix, so CI here will be red until #163 lands — a green tick bought by bundling someone else's change is worth less than a PR that reviews in ten seconds. Once #163 merges, this makes `env_var_registry` pass.

### Bearing on the other open PRs

aionhw/xezim#165 is currently red for three independent reasons, and only one of them is that PR's own: its regression tests fail by design until the core `rev` bump from aionhw/xezim-core#44, plus the two `main` problems above. Worth knowing before reading its CI.


aionhw (approved):
Confirmed with #163 applied: env_var_registry fails on main with exactly this ordering diff and passes with the line moved. Merging.

Issue #45, opened 2026-07-22
Clocking Blocks Support


Ganesh-TNL:
Sample test on EDA Playground for various scenarios involving clocking blocks

https://www.edaplayground.com/x/iAUb

<details>

<summary>SystemVerilog Code for Clocking Blocks Test Case</summary>

```systemverilog
`timescale 1ns/1ps

// ============================================================================
// 1. SELF-CHECKING PREAMBLE MACROS
// ============================================================================
`ifndef SVTEST_DEFS_SVH
`define SVTEST_DEFS_SVH

`define SVTEST_INIT \
  int failures = 0;

`define SVTEST_CHECK(expr, msg) \
  if (!(expr)) begin \
    failures++; \
    $display("FAIL: %s at time %0t ps", msg, $time); \
  end

`define SVTEST_PASSFAIL \
  if (failures == 0) begin \
    $display("TEST_PASS"); \
  end else begin \
    $display("TEST_FAIL count=%0d", failures); \
    $fatal(1); \
  end

`endif

// ============================================================================
// 2. TARGET DUAL-DOMAIN INTERFACE DESIGN (Sec 14.3)
// ============================================================================
interface bus_if(input bit clk);
  logic [7:0] data_bus;
  logic       cmd_valid;
  logic       ready_b;

  // Use-Case A: Standard Default Clocking Block with input/output skews
  // Input sampled 2ns before edge; Output driven 1ns after edge (Sec 14.4)
  default clocking cb_main @(posedge clk);
    default input #2ns output #1ns;
    input  data_bus;
    output cmd_valid;
  endclocking

  // Driver clocking block to isolate testbench stimulus drives from sampling registers
  clocking cb_drv @(posedge clk);
    default output #1ns;
    output data_bus;
  endclocking

  // Use-Case B: Multi-clocking / Alternate Skew Definition Block
  clocking cb_fast @(negedge clk);
    input  data_bus;
    output #0 ready_b; 
  endclocking
endinterface

// ============================================================================
// 3. MAIN REGRESSION TESTBENCH MODULE
// ============================================================================
module tb_ieee1800_clocking_blocks;

  `SVTEST_INIT

  // Infrastructure Clock Generators
  bit clk_sys;
  always #5ns clk_sys = ~clk_sys; // 100MHz master clock reference (Period = 10ns)

  // Instantiate Interface Under Test
  bus_if master_if(.clk(clk_sys));

  // Initialize raw wire states safely at time 0
  initial begin
[...]

PR #168, opened 2026-09-11
Regression tests: a nettype net declared in a sub-module keeps its declaration

vrajeshprakhya:
Regression tests for aionhw/xezim-core#45, which fixed a continuous assign to a user-defined nettype net declared inside a **sub-module** being silently dropped: the net read its members' defaults, with no error, and the same declaration in the top module was correct.

```systemverilog
module mid import p::*; ();
  nd dd;                    // nd: a nettype over an unpacked struct (§6.6.7(d))
  assign dd = dv(3.3);
  initial $display("SUB=%0.4f $bits=%0d", dd.v, $bits(dd));
endmodule
module tb; mid u_mid (); endmodule
```

printed `SUB=0.0000 $bits=1` with a §6.10 implicit-net warning for `u_mid.dd`; the same body as the top module printed `SUB=3.3000 $bits=129`.

## The three cases

* **the net keeps its value, and draws no implicit-net warning.** `$bits` is asserted too — 1 is the signature of the declaration having been replaced by an implicit scalar, and it distinguishes "shadowed" from "resolved to the wrong number".
* **adding a second uninstantiated module does not change the answer.** That changes which module root auto-detection picks, demoting the one under test to a sub-module — which is how a working design broke by having an unrelated module added beside it. Asserted as an equality between the two runs, so it pins the *independence*, not a particular value.
* **the net still goes through its resolution function.** Two current drivers sum and a single voltage owner establishes the voltage. The declaration surviving is not sufficient on its own: a net that is present but never resolved is the same wrong answer wearing a declaration, and a fix for the shadowing alone produces exactly that.

All three fail on the unfixed core and pass with it — verified by reverting the fix, rebuilding, and re-running.

## Note on the first commit

This PR includes the one-line fix from #167, without which the `misc` group does not compile on `main` (a raw string in `elaboration_shape_fixes.rs` had its inner quotes re-escaped by 2486c17). If #167 merges first, that commit drops out cleanly on rebase.

https://claude.ai/code/session_014eLBkW9EERLgZ9jU8TcRSg

### DIFF
--- a/tests/misc.rs
+++ b/tests/misc.rs
@@ -343,3 +343,5 @@ mod while_continue_final_iter;
 mod stdout_flush_at_sim_finish;
 #[path = "misc/elaboration_shape_fixes.rs"]
 mod elaboration_shape_fixes;
+#[path = "misc/nettype_net_in_submodule.rs"]
+mod nettype_net_in_submodule;
--- a/tests/misc/nettype_net_in_submodule.rs
+++ b/tests/misc/nettype_net_in_submodule.rs
@@ -1,0 +1,163 @@
+//! IEEE 1800-2017 §6.6.7(d), §7.2, §6.10: a user-defined nettype whose data
+//! type is an unpacked struct, declared as a net inside a SUB-MODULE.
+//!
+//! An unpacked struct has no signal under its own name — elaboration registers
+//! the member LEAVES (`base.i`, `base.v`, ...) and nothing at `base` — which is
+//! also how a nettype net of that type is stored. The post-inlining implicit-net
+//! pass looked for a signal at the base name, did not find one, and laid a 1-bit
+//! implicit net over it (§6.10). That scalar then absorbed the continuous
+//! assign: the member leaves were never written, and every read of the net came
+//! back 0.0.
+//!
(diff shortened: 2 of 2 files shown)


aionhw (approved):
Reviewed together with aionhw/xezim-core#45.

Verified here on today's main (xezim 6d2d0b2, core a33f3f9):
- the three tests fail on current core exactly as described (all three, not just the first), and pass with core#45 merged onto core main;
- the reference simulator agrees with every expected value: `SUB=3.3000`, `$bits=129`, and `I=4.0000 V=1.2500` for the three-driver case;
- misc, hierarchy, types, scheduling and classes groups plus the core suite are green with core#45 merged.

The tests themselves are well chosen — the `$bits == 129` check and the root-independence equality are exactly the two signatures that separate "shadowed by an implicit net" from "resolved wrong", and the three-driver case pins the resolver path that a shadowing-only fix would miss.

Merge order: core#45 first, then bump the `rev` pin in `Cargo.toml` (both lines, see the comment above them) in the same push that lands this, otherwise a fresh clone builds against a core where these three tests fail.

PR #169, opened 2026-09-11
sim: collection == / sequence-item binding / undefined-method diagnostics on parameterized classes

Thomasb81:
Three fixes to parameterized-class semantics and diagnostics, each with a regression self-test.

## Commits
1. **Report calls to undefined subroutines inside class methods** (`3fcd552`)
   Emit a diagnostic when a class method calls a subroutine that does not exist, instead of silently ignoring it; module-level task calls still work.

2. **Fix sequence-item type resolution from the owning instance's class chain** (`c02eb16`)
   A parameterized base's type-param field accessed through a foreign specialization must return the owner's own binding.

3. **Fix collection `==` to compare element values, not `Value` structs** (`da6c5e2`)
   `compare_dyn_arrays` used structural equality (including width); it now uses semantic element equality (`is_equal`, zero-extending the narrower unsigned operand, IEEE 1800-2017 §11.6.1).

## Testing
- New self-tests pass: `undef_method_call_reported`, `param_instance_field_uses_owner_binding`, `queue_dynarray_eq_inplace_overwrite`.
- Verified byte-for-byte against reference simulators.

## Note on coordination with #167
The `tests/misc/elaboration_shape_fixes.rs` string-escape regression is **separately fixed by PR #167** (and by a raw-string variant in #163). To avoid a conflicting duplicate, this PR intentionally does **not** change that test file and defers to #167 for it. #167's touched files (`tests/misc/...`, `tests/scheduling/...`) share **zero overlap** with this PR's files, so the two merge independently without conflict.

### DIFF
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -100852,6 +100852,37 @@ impl Simulator {
                 self.exec_task_call(&td, args);
                 return Value::zero(32);
             }
+            // An unqualified call inside a class method that resolved to NO
+            // method of the class (or its ancestors), NO package or module
+            // subroutine, and NO let — i.e. the call is a genuine undefined
+            // subroutine reference. Per IEEE 1800-2017 §13.3, a call to an
+            // undeclared task/function is an error; the reference simulator
+            // rejects it at elaboration ("Failed to find 'X' in hierarchical
+            // name"). xezim used to silently drop these as a no-op returning
+            // zero, so stale UVM-1.x calls like `kill_child_sequences()`/
+            // `kill_sequence_activity()` (removed from the 1800.2 library)
+            // ran without complaint and the surrounding logic kept executing
+            // on garbage — e.g. a sequence-abort regression no longer registered the
+            // still-pending child process, masking the real failure. Now we
+            // report it and halt, matching the reference's stance.
+            //
+            // Guard: only in a class-method context (`class_context_stack`
+            // top is `Some(Some(class))`). Top-level global function calls
+            // (e.g. `get_finish_on_completion()` at module scope) also reach
+            // this fall-through for UVM global helpers with no body in scope;
+            // the reference tolerates those too (they bind to uvm_root), so
+            // we must NOT error there — a full-suite run confirmed that
+            // restricting the error to class-method context flags exactly one
+            // test and no currently-passing one.
+            if let Some(Some(ctx)) = self.class_context_stack.last().cloned() {
+                eprintln!(
+                    "[xezim][error] call to undefined subroutine '{}' inside class '{}' \
+                     (no such method or in-scope task/function); reference simulators reject \
+                     this at elaboration",
+                    name, ctx
+                );
+                std::process::exit(1);
+            }
         }
         Value::zero(32)
     }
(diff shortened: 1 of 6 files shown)


aionhw (changes requested):
Thanks — the three fixes are well isolated and each comes with a regression test. Commits 2 (sequence-item binding order) and 3 (collection `==` by element value) look right to me and I would take them as they are. Commit 1 needs another pass before it can go in, because on this branch merged with current main two of the six UVM AVIP regressions abort:

```
[xezim][error] call to undefined subroutine 'ifSignalsAreStable' inside class 'uvm_root' (no such method or in-scope task/function); ...
[xezim][error] call to undefined subroutine 'ifValidHighThenInformationAreNotUnknown' inside class 'uvm_root' ...
```

(spi_avip and axi4Lite_avip; both exit 1 with no UVM summary; axi4, apb, i3c, ahb still pass.) Both names are SVA `property` declarations — `property ifValidHighThenInformationAreNotUnknown(logic valid, logic information); ... endproperty`, instantiated by `assert property (ifValidHighThenInformationAreNotUnknown(arvalid, araddr))` in `Axi4LiteSlaveReadAssertions.sv` (and the spi one in `SpiSlaveAssertions.sv`). So two things are off in the guard:

1. The unresolved-call fall-through is reached by more than subroutine calls: a property/sequence instance evaluated by the assertion machinery lands there too. The check needs to exclude names bound as properties/sequences (and, defensively, anything the elaborator knows as a non-subroutine).
2. `class_context_stack` top being `Some(Some(class))` does not mean the call is lexically inside a class method. Here the assertions live in an interface, but they are evaluated while `uvm_root`'s `run_phase` is the innermost class context on the stack, so the message even names the wrong owner. The context that matters is the lexical scope of the call site, which the interpreter has (the statement's scope / declaring class), not the dynamic stack.

Separately from the guard, `std::process::exit(1)` from inside the simulator skips everything a fatal is supposed to do: the exit-code contract (`tests/misc/exit_codes.rs`), `$finish` bookkeeping, VCD/FST finalization, buffered stdout, and it kills the host when xezim runs as a library (`xezim::simulate()` in the test suite, cocotb, VPI hosts). The existing precedent for a hard exit is an elaboration-time allocation guard in `Simulator::new`; runtime fatals go through `$fatal`'s path (`self.fatal_finish_number = Some(1); self.finished = true;`), and that is what I'd use here. Better still would be reporting this at elaboration in core's validator, next to the existing declared-nowhere check for module bodies, since the reference simulators reject it there too.

Small note on the tests: `undef_method_call_reported.rs` returns silently when the `xezim` binary is not built, which makes it pass vacuously under `cargo test --lib`-style runs. That matches the pattern already used in `tests/scheduling/determinism_and_stall.rs`, so no change requested — just flagging that it is a known vacuous-pass shape.

What I ran: branch merged with main (77 commits behind at review time, clean merge), release build, c906 hello bit-exact, the six AVIPs (spi and axi4Lite fail as above), and the three new tests plus the classes (351), collections (136) and misc (851) groups on the merged build — all pass.

aionhw:
Closing: the review found the undefined-subroutine diagnostic aborting the spi/axi4Lite AVIPs on SVA property names and the process::exit path (see the CHANGES_REQUESTED review). Happy to look at a reopened PR that addresses those.

PR #172, opened 2026-09-15
feat: --stems RTLBrowse sidecar for source-annotated tracing

opensource-elearning:
## Summary

Adds a `--stems <file>` option that writes an RTLBrowse `.stems` companion
file next to a waveform dump. GTKWave's `rtlbrowse` view then opens the
design source and annotates live signal values onto the lines while the
waveform plays:

```sh
xezim --simulate --fst wave.fst --stems out.stems counter.sv
gtkwave -t out.stems wave.fst
```

Until now that workflow needed a converter step (`xml2stems`) driven from
another tool's output. xezim now writes the same record format directly from
its own elaborated module, so the companion file comes out of a normal run.

## The file

One record per line, two kinds:

```text
++ module <name> file <path> lines <start> - <end>
++ comp <instance> type <module-type> parent <parent-module-name>
```

- `++ module` names every module definition in the design — the top module
  plus each instantiated definition — its source file, and the line its
  header keyword starts on.
- `++ comp` turns the instance tree into the hierarchy RTLBrowse renders:
  each instance, the definition it instantiates, and the module that
  contains it. A module no `++ comp ... type` record references is a tree
  root.

Declaration lines come from a header-keyword scan of the preprocessed source
text, the same text elaboration uses. Interfaces, programs, and macromodules
resolve like modules, so instances of any of them land in the tree. Comments
and attributes above a header are blanked to spaces before the scan and never
shift the reported line.

## Behavior on edge cases

- A compiled-artifact run (`--compile` then `--simulate`) carries no
  preprocessed source text: declaration lines fall back to 1, file paths
  stay real.
- The stems grammar tokenizes fields with `%s`, so a source path containing
  whitespace cannot be represented. Paths pass through verbatim, and spaces
  in paths are the author's to avoid — the same rule `xml2stems` imposes.
- An escaped identifier in a `module` header falls back to line 1.
- A failed stems write warns to stderr and does not disturb the simulation,
  matching FST-dump failure behavior.
- A module defined inside a file pulled in by `` \`include `` is attributed to
  the including file at the spliced position. The records stay structurally
  valid (hierarchy and parents all resolve), but the file/line pair for such
  a module can point at a blank or spliced line. Keeping module definitions
  out of include files avoids the ambiguity until the preprocessor tracks
  include provenance.

## Changes

| File | Change |
| --- | --- |
| `src/compiler/act_stems.rs` | new — module-site resolution and the stems writer |
| `tests/act_stems.rs` | new — 21 integration tests |
| `src/compiler/simulator.rs` | `set_stems_cli` process-global, `Simulator::stems_file`, sidecar written at `compile()` dump setup |
| `src/main.rs` | `--stems` / `--stems=` parsing at both `Simulator::new` sites, help text |
| `src/compiler/mod.rs` | module export |
| `README.md` | option table row + feature bullet |
| `docs/source-annotation.md` | new — worked example with real output, format reference, edge cases, Surfer status |

## Tests

All 21 tests spawn the real binary against scratch designs and parse the
generated stems back with an oracle that models the RTLBrowse grammar, so the
assertions are independent of the writer.

| Case | Coverage |
| --- | --- |
| Hierarchy / flat / multi-file | 3-level instance tree, single module no comps, per-module source files |
| Repeated instantiation | one `++ module` record per definition, comp parent linkage kept even with colliding instance names |
| Deep chains, forward refs, interfaces | intermediate parent resolution, line-exact forward references, interface instances |
| Header forms | module, ANSI- and parameterized-header scan, comments/attributes above a header, escaped header fallback, macromodule |
| Negative | malformed stems rejected by the parser test; elaboration failure writes no sidecar |
| Corner | artifact load keeps paths but falls back to line 1; whitespace-in-path pass-through pinned; include-spliced modules attributed to includer; bad path warns & continues; `--stems=` form parity |

## Verification

- `cargo test --test act_stems` — 21 passed, 0 failed (bytecode mode).
- `cargo test --features jit --test act_stems` — 21 passed, 0 failed.
- `rustfmt --check` clean on the two new files; `cargo clippy` reports no
  diagnostics on the new or modified code.

## Surfer

Surfer reads the FST the same run produces, so the artifacts open there
today. Surfer has no source window yet and does not read `.stems`; the
sidecar plus the FST's scope hierarchy are the open contract such a window
would consume, documented in `docs/source-annotation.md`.

cc: @aionhw , @oscargus , @gtkwave, @rfuest

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ fst-reader = "0.17"
 # check that its header declares the compression (XTrace §6.8). Same version
 # xezim-core writes the stream with.
 zstd = "0.13"
+tempfile = "3.14"
 
 [build-dependencies]
 cc = "1"
--- a/src/compiler/simulator.rs
+++ b/src/compiler/simulator.rs
@@ -8194,8 +8194,33 @@ impl Simulator {
     /// Configure the worker-thread count. `n >= 2` enables the background
     /// VCD writer thread; `n == 1` keeps the dump path inline.
     pub fn set_threads(&mut self, n: usize) {
-        self.threads = n.max(1);
+        self.threads = n;
         self.pdes_worker_pool = None;
+    }
+
+    /// Centralized worker cap for internal dispatch paths.
+    ///
+    /// The CLI `--threads` value (stored in `self.threads`) is the user's
+    /// requested thread count. The internal dispatch caps at 8 workers based
+    /// on measured performance (persistent-worker path measured slower than
+    /// the scoped path on c910), but never exceeds the user's request.
+    ///
+    /// Returns the effective worker count for dispatch: min(user_request, 8),
+    /// with a floor of 1. On machines where available_parallelism() fails,
+    /// defaults to 2 workers (preserving the previous .unwrap_or(2) behavior).
+    fn safe_worker_cap(&self, bound: usize) -> usize {
+        let avail = std::thread::available_parallelism()
+            .map(|n| n.get())
+            .unwrap_or(2);
+        self.threads.min(avail).min(8).min(bound).max(1)
+    }
+
+    /// Variant for callers without a natural bound (e.g., BSP settle).
+    fn safe_worker_cap_unbounded(&self) -> usize {
+        let avail = std::thread::available_parallelism()
+            .map(|n| n.get())
+            .unwrap_or(2);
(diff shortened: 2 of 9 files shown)


opensource-elearning:
Superseded by #173 with clean single-commit history

Issue #10, opened 2026-07-09
%p Assignment Pattern Format Specifier Behavior


Ganesh-TNL:
Sample test case on EDA Playground:

https://www.edaplayground.com/x/UXU8

Tagged unions are purposefully avoided in the test case above, as they are not universally supported by Aldec + Big 3 simulators.

Simulation log with Aldec Riviera Pro in the above link shows %p specifier obeying LRM mandates (Section 21.2.1.7 of the IEEE Std 1800-2017 : Assignment pattern format). 

From my use-case perspective, the important aspect from the LRM specs is the **named elements** for structs / class objects.

Output from xezim 0.8.2: (Related note - can the first line === xezim === include the version number as well?)

```
ganesh@DEV-MC-ADL-S:~/foss/sv-test-cases$ ../xezim/target/release/xezim percent_p_format.sv
=== xezim ===
Max time: 100000
------------------------------
[PHASE] compilation: 1.3ms
[EVENT-EDGE] measure (timestamp): 0 edge blocks, 0 gateable, 0 gateable-with-EMPTY-data-reads, avg data-reads/gateable=0.00
[EVENT-EDGE] SKIP mode ON (sim_time >= 0) — skipping gateable flop fires with no data-input change
=== STARTING IEEE 1800-2017 DEEP NESTED & CHARACTER LIMIT TEST ===

[SEGMENT 12 - Deep Nested Aggregate Evaluation]
Description: Printing a structure containing arrays of structures with inner unions, packed items, enums, and strings.
Actual %p Output:
'{999, 0.945, x}

[SEGMENT 13 - 1024 Character Minimum Limit Test]
Description: Printing an exceptionally long unpacked array (60 records deep) to check for truncation or compiler warnings.
Note: If the simulator supports/reaches its internal limit, look for a trailing ellipses (...) or an explicit truncation warning below.
Actual %p Output:
'{x, 0.0, x, x}

=== DEEP NESTED & CHARACTER LIMIT TEST COMPLETE ===
=== LRM EXAMPLE TEST ===
cmb[int] = '{x, x};
cmb[int] = '{x, x};
cmb[10].str = 0;
=== LRM EXAMPLE TEST COMPLETE ===
[PROF] settle=0.0ms edges=0.0ms nba=0.0ms process=1.7ms snap=0.0ms sched=0.0ms
[PROF] edge_waiters=0.0ms edge_cg=0.0ms waiter_iters=0
[...]

PR #174, opened 2026-09-15
Honor loop break/continue before a call to a blocking subroutine

chenbhil:
## What

A `break` or `continue` inside a loop is dropped when a later statement in the same iteration calls a subroutine that blocks (a task with `#delay`/`@event`/`wait`). The guarded iterations run the blocking call anyway.

```systemverilog
for (int i = 0; i < N; i++) begin
  if (skip(i)) continue;   // ignored
  do_work(i);              // a task that blocks; runs even when skip(i)
end
```

Inline delays are fine. Only a blocking edge reached through a call is affected. `continue` is dropped entirely; `break` fires one iteration late.

## Root cause

In `simulator.rs::run_process_stmts`, blocking statements (a call to a time-consuming task/method, `#delay`, `@event`, `fork`, blocking begin/end) are intercepted ahead of `exec_statement`. The top guard in `exec_statement` is what no-ops each statement while `break_flag`/`continue_flag` is pending, so the blocking path never sees it. Inlining a blocking task compounds the problem: `bind_task_frame` saves and clears those flags for the callee body.

## Fix

One guard at the top of the statement loop, mirroring `exec_statement`. While a flag is pending (and `disable_target` is None), skip the current statement unless it consumes the flag: the loop re-entry tails, `LoopStep`, and `ScopePop`.

## Tests

Regression tests in `tests/scheduling/blocking_loop_break_continue.rs`: `blocking_for_continue_skips_call_to_blocking_task` and `blocking_for_break_skips_call_to_blocking_task`. The pre-existing cases put the `#delay` before the jump and never covered this shape. Full suite: 2599 passed, 12 failed (the known macOS VPI-linker family; the +2 are the new tests). Zero regressions.


aionhw (changes requested):
Thanks — the diagnosis is right (a blocking call is intercepted in `run_process_stmts` ahead of `exec_statement`, and `bind_task_frame` clears the flags), and both new tests pass here, as does the full `scheduling` group.

The guard has one hole, though: the "flag consumer" exemption keys on the statement *kind*, but a **fresh nested loop** that simply follows the `break`/`continue` has the same kind as the re-entry tail of the loop being iterated. Such a loop is let through, its entry gate (`blocking_loop_flag_gate`) consumes the *outer* loop's flag as if it were its own, and the nested loop body runs. So `continue` followed by a nested `for`/`while`/`repeat` (all with a blocking call inside) is still ignored, and `break` followed by a nested `foreach` is turned into "skip this iteration" — the outer loop keeps going.

Probe on this branch (each nested loop calls a task with a `#10` inside):

```systemverilog
task automatic run_nested();          // continue, then nested for
  for (int i = 0; i < 8; i++) begin
    if ((i % 2) == 0) continue;
    for (int j = 0; j < 2; j++) begin inner_runs++; work(i); end
  end
endtask
task automatic run_break_nested();    // break, then nested foreach
  for (int i = 0; i < 8; i++) begin
    outer_iters++;
    if (i >= 2) break;
    foreach (hits[k]) if (k == 0) work(i);
  end
endtask
task automatic run_repeat();          // continue, then nested repeat
  for (int i = 0; i < 4; i++) begin
    if (i != 1) continue;
    repeat (1) begin work(i); r_runs++; end
  end
endtask
```

Results with this PR's build (expected: `inner_runs=8`, `outer_iters=3`, `r_runs=1`):

```
NESTED inner_runs=16 hits='{2, 2, 2, 2, 2, 2, 2, 2}   // continue ignored for even i
BREAK  outer_iters=8                                  // break consumed by the foreach; loop ran to 8
REPEAT r_runs=4 hits='{1, 1, 1, 1, 0, 0, 0, 0}         // continue ignored
```

(`while` after `continue` fails the same way; `do…while` and a `fork` after `continue` are correct.) This is the same shape as the FlooNOC case whenever the guarded code after `continue` is itself a loop over neighbours/flits, so I think it is worth closing here rather than in a follow-up.

Suggested direction: only a loop statement that is a **re-entry** of the loop currently being iterated should be allowed to consume the flag; a loop statement reached for the first time while `break_flag`/`continue_flag` is set must be skipped like any other statement. The re-entry sites are the places that push `stmt.clone()` as the continuation tail (`pc.pushed(...)`), so the tail copy can be marked there (a wrapper kind or a per-process "re-entering loop span" field checked in the new guard) — `ForeachTail`/`ForeverTail` already are distinguishable, and the `LoopStep` barrier is fine as is.

Once that is in, please add the nested-loop shapes above to `blocking_loop_break_continue.rs` (they are the cases the current tests can't tell apart).

aionhw:
Closing: the guard still lets a fresh nested loop after continue/break consume the flag (details and probe in the review). A reopened PR that marks loop re-entry tails, with the nested-loop shapes added to blocking_loop_break_continue.rs, would be welcome.

chenbhil:
Reproduced. found out that  a fresh nested loop got treated as if it were consuming its own enclosing loop's flag.
It will take some time. Will work on that next week.

PR #46, opened 2026-07-22
Add CI and clippy cleanups

oscargus:
This will show that some of the current tests fail, but so they do on current main.

```
failures:
    blocking_for_honours_break
    blocking_repeat_honours_break
    blocking_repeat_honours_continue
    blocking_while_honours_break_and_continue
```

PR #175, opened 2026-09-16
fix: wide value parsing in $fscanf/$sscanf/$value$plusargs and string methods

Ganesh-TNL:
## What

Three integration points in `simulator.rs` parse decimal/hex/octal/binary
strings with hard-coded 64-bit assumptions, producing wrong results for wide
destination variables (e.g. `logic [159:0]`):

1. **`$sscanf` / `$fscanf`** (`sv_sscanf_consumed`): parses the numeric token
   with `i64::from_str_radix(text, radix)`, then assigns via
   `Value::from_u64(n as u64, 32)`. For tokens >= 2^63 the `i64` parse overflows
   → returns `Err` → breaks the scan loop → `cnt=0`, file position **not
   advanced**. Production symptom: infinite re-read of the same token → "wrong
   cmd" fatal.

2. **`$value$plusargs`** (`parse_plusarg_value`): `%d` uses
   `cleaned.parse::<i64>()` → `None` for values >= 2^63 → plusarg reported as
   **not found**. `%h`/`%o`/`%b` hard-code `width=64` regardless of the
   destination signal width → wide values are silently truncated to 64 bits.

3. **`eval_value_plusargs`**: does not pass the destination width to
   `parse_plusarg_value`, so even after fixing (2), width information is lost.

Affects: `$sscanf("deadbeefdeadbeef…", "%x", w160)`,
`$fscanf(fd, "%d", w160)` with a 20-digit seed,
`$value$plusargs("SEED=%d", w160)` with `+SEED=9223372036854775808`.

## Root cause

```rust
// simulator.rs — sv_sscanf_consumed (line ~10554):
match i64::from_str_radix(&text, radix) {
    Ok(mut n) => { /* assigns as 32-bit regardless of dest width */ }
    Err(_) => break,   // ← 20-digit decimal overflows i64 → breaks scan
}

// simulator.rs — parse_plusarg_value (line ~10336):
'd' => cleaned.parse::<i64>() …      // fails for >= 2^63
'h' | 'x' => Value::from_str_radix(s, 16, 64)  // hard-coded 64
'o' => Value::from_str_radix(&cleaned, 8, 64)  // ditto
'b' => Value::from_str_radix(&cleaned, 2, 64)  // ditto
```

### 1. `parse_plusarg_value(raw, spec, width)` — new `width` parameter

- `%d`: arbitrary-precision decimal via `u128` accumulator for widths <= 128;
  falls through to `Value::from_str_radix(text, 10, w)` (which now handles wide
  decimal via the companion xezim-core PR) for wider destinations.
- `%h`/`%x`/`%o`/`%b`: passes `width` to `from_str_radix` instead of 64.

### 2. `eval_value_plusargs` — infer destination width

```rust
let dest_width = self.infer_lhs_width(&args[1]).max(1);
// … passed into parse_plusarg_value(suffix, spec, dest_width)
```

### 3. `sv_sscanf_consumed` — width-aware parsing + two's complement

```rust
let dest_width = if oi < outs.len() {
    self.infer_lhs_width(&outs[oi]).max(1)
} else { 32 };
let mut v = Value::from_str_radix(&text, radix, dest_width);
if neg {
    v = v.bitwise_not().add(&Value::from_u64(1, dest_width));
    v.is_signed = true;
}
```

Replaces the `i64::from_str_radix` → `Ok/Err` path with the core's now-
width-aware `from_str_radix`. Negative `%d` applies two's complement via
`bitwise_not().add(1)` at the destination width rather than `i64::neg`.

## Tests

Regression tests: `tests/strings/wide_string2num.rs` (7 test functions, 50+
assertions). Registered in `tests/strings.rs`.

| Test | LRM | Covers |
|------|-----|--------|
| `sscanf_wide_hex_preserves_full_width` | §11.4.12 | 40-hex → 160-bit, overflow wrap to 32-bit |
| `sscanf_wide_bin_and_oct` | §11.4.12 | 80-bit binary, octal >= 2^63 → 160-bit |
| `sscanf_wide_decimal_signed` | §11.4.12 | 20-digit +/− decimal → 64-bit |
| `sscanf_field_width_16x` | §11.4.12 | `%16x` field-width with wide dest |
| `fscanf_wide_advances_past_large_token` | §11.4.12 | File position advances; clean EOF = −1 |
| `string_atoi_wide_wraps_to_low32` | §6.16.8 | 50/20-digit decimal → int |
| `string_atohex_atooct_atobin_wide` | §6.16.8 | wide hex/oct/bin → int; atoreal |

All 7 fail on unfixed code (values = 0, scan breaks). All pass with this change.

## Verification

```
cargo test --release --lib                           →  39 passed
cargo test --release --test strings                  → 135 passed  (7 new)
cargo test --release --test scheduling               → 348 passed
cargo test --release --test classes                  → 340 passed
cargo test --release --test types                    → 437 passed
cargo test --release --test hierarchy                → 184 passed
cargo test --release --test gates                    →  97 passed
cargo test --release --test collections              → 131 passed
cargo test --release --test perf                     →  22 passed
cargo test --release --test upf                      →   4 passed
─────────────────────────────────────────────────────────────────
Total                                                → 1737 passed, 0 failed

MWE data_parsing_via_fmt_spec (production TB)        → TEST_PASS
```

Note: `tests/misc` has a pre-existing compile error (`expected expression,
found %` in `elaboration_shape_fixes.rs`), confirmed present on clean checkout.

## Dependency

Needs `aionhw/xezim-core` PR
(`feat: Arbitrary precision decimal parsing for wide values in from_str_radix`): the core 
`from_str_radix(…, 10, width)` path must handle wide
decimal or the `%d`/`%d` scan paths fall through to 0 for values >= 2^64 even
with the correct width propagated.

## Files changed

- `src/compiler/simulator.rs` (+51, −24)
- `tests/strings/wide_string2num.rs` (new, 204 lines)
- `tests/strings.rs` (+2)


aionhw (changes requested):
Thanks — routing `$sscanf`/`$fscanf`/`$value$plusargs` through `Value::from_str_radix` with the destination width is the right shape, and together with aionhw/xezim-core#46 it also fixes wide *decimal literals* (`128'd340282366920938463463374607431768211455` currently elaborates to 0 on main), which is a bigger win than the title says.

Two things to fix before merge, one to tidy:

1. **`$value$plusargs("N=%d", n)` lost the sign.** The old path parsed with `i64` so `+N=-5` gave `-5`; the new `'d'` arm folds `cleaned.chars()` through `to_digit(10)`, which silently drops the `-` (and any other non-digit), so `+N=-5` now assigns `5`, and `+N=abc` now "succeeds" with `0` where it used to return 0/not assign. The `$sscanf` arm already handles the sign separately (`neg` + two's complement) — please do the same here (strip a leading `-`/`+`, then negate at `w` bits), and keep returning `None` when no digit was consumed.
   Probe: ``+N=-5` (int n) → main prints `N=-5`, this branch prints `N=5`; `+BAD=abc` (int u) → main "rejected", this branch assigns 0 and returns 1.`

2. **Dependency pin.** This branch needs the decimal-wide branch of `from_str_radix` from xezim-core#46 for anything above 128 bits (`from_str_radix(&cleaned, 10, w)` is exactly the call that returns zero on today's pinned core), but `Cargo.toml` still pins core `89678c7`. Merge order has to be core#46 first, then bump the `rev` here in the same commit (see the comment above the pin) — otherwise a fresh clone of main builds the old behaviour and the >128-bit path silently returns 0.

3. Tidy: `fscanf_wide_advances_past_large_token` writes `wide_fscanf_test.txt` into the crate root; a unique per-test name (or `std::env::temp_dir()`) keeps it from colliding with the other file-based tests when the suite runs in parallel (we had a flaky case of this in #164).

Core side (xezim-core#46), reviewed there: the limb loop is correct, but `from_decimal_str` ignores non-digit characters instead of rejecting them and is mis-indented relative to the `impl` block; comments on that PR.

Suite: built against core#46 via a path patch; the new `wide_string2num` tests and the whole `strings` group pass (135/135).

aionhw:
Closing: $value$plusargs %d loses the sign and accepts non-numeric text, and the core pin needs bumping after xezim-core#46 lands (see the review). Please reopen with those addressed.

